PR #2, opened 2017-10-06
Fix for missing public folder

jsonmaur:
Resolving issue https://github.com/nozzle/react-static/issues/1.


derekperkins:
Thanks!!

PR #23, opened 2017-10-10
README: Fix link to #configuration-staticconfigjs

chadoh:
Links was missing a `#` at the beginning

PR #33, opened 2017-10-12
Fix README link to project setup

calvinrbnspiess:
Just the same missing `#` for the `Project setup` section like in #23


derekperkins:
Thanks again!

PR #62, opened 2017-10-19
Fix .gitignore file not being copied over from the template

Maxwell2022:
Solution based on `create-react-app`


tannerlinsley:
Genius

D1no:
@Maxwell2022 I just created an example based on styled-components with `3.0.6` and don't have a `.gitignore` in my folder.
image

Issue #73, opened 2017-10-22
add sass to webpack config


tzookb:
Read the docs of how to add rules to webpack,
so I did this simple thing:

```
function withSass(config) {
  const sassIt = {
    test: /\.scss$/,
    loader: "css-loader!sass-loader"
  };
  config.module.rules.push(sassIt);
  return config;
}



and on the config
{
    ...
    webpack: [withCss, withFiles, withSass],
}
```

but style file doesnt get created... what am I doing wrong?

tzookb:
this is what I ended up doing, quite hacky but works :/

```
function withSass(config) {
  const sassIt = {
    loader: 'sass-loader',
    options: {
      noIeCompat: true
    }
  };

  config.module.rules[1].test = /\.scss$/
  config.module.rules[1].use.push(sassIt)

  return config;
}
```

Issue #70, opened 2017-10-21
Importing typefaces


calvinrbnspiess:
Although importing CSS is easy with the new version `2.0.0`, it is still not possible to import fonts out-of-the-box. #40 

The following could easily remedy the situation:

1. Adding font file extensions like `.woff`, `.woff2`, `.eot`, `.ttf` to the `withFileLoader`
2. Making the `withFileLoader` independent from the file extension by removing the test expression and by providing an exclude for `.css` and `.js` files
3. Providing a plugin out-of-the-box to load fonts e.g. `withFontsLoader`:
```javascript
const withFontsLoader = config => {
	config.module.rules = config.module.rules.concat([
		{
			loader: 'url-loader',
			test: /\.(png|woff|woff2|eot|ttf|svg)$/,
			query: {
				limit: 10000,
				name: 'static/[name].[hash:8].[ext]',
			},
		},
	]);
	return config;
};
```

Thanks!

tannerlinsley:
Would you like to make a PR to the plugins directory? It could include this
and also update the other file loader plugin with the changes you
mentioned.

D1no:
Funny how we are getting back to gatsbys plugin system 😅 I think requiring fonts is probably something that should also work "out of the box"

alexiscordova:
I don't necessarily see this as a path towards a react-static-plugin system (these are just webpack rules, after all), but it could be tough to constantly update the `withFileLoader` plugin with everything that should migrate to `dist`.

I think the default option should be one of the following:
- Anything that is not supposed to be compiled (images, fonts, etc.) in `src` will move to `dist` (essentially following the 2nd option @calvinrbnspiess suggested)
- Standard image filetypes in `src` will move to `dist`, but anything else has to be a custom loader.

I'm personally good with either, since webpack rules are relatively straightforward to implement, but the documentation should be updated to reflect what is and is not supported by default.

calvinrbnspiess:
@tannerlinsley I think @alexiscordova and @D1no's arguments are really good and I also think that the most important file types should be requirable without any further configuration.
I will propose a PR that allows us to load the most important file types via file-loader, but allows us to opt-out and provide custom loaders via those webpack transformer functions.

PR #75, opened 2017-10-22
Update webpack module rules to allow importing everything that is not supposed to be compiled

calvinrbnspiess:
**TL;DR** Solves #70.  Probably makes things like #73 and configuring webpack loaders easier. Working out-of-the-box without any configuration in most cases.

---

Basically, I decided to change the current behaviour a bit. I dropped the newly introduced `./plugins` directory in favor of `./webpack/rules` where I created modules to be able to require every loader seperately. I think the `withCSSLoader` or `withFileLoader` did not fulfilled the purpose of plugins and were way too simple to be required even by ordinary tasks like loading `.css` files.

The main cause for this pull request was to address my personal issues with loading fonts via the `typefaces`-packages. Actually, it should not be neccessary to configure loaders for every file-type.

> Funny how we are getting back to gatsbys plugin system 😅 I think requiring fonts is probably something that should also work "out of the box".

Now everything that is not supposed to be compiled can be imported too. Basically this approach is the most struggle-free way to solve this issue and this solution won't affect the bundle size at all, but afterall it inflates our toolchain a bit and simultaneously takes some effort from someone who just want to use this `static-site-generator` without digging through documentation or configuration files.

The included things are basically:
```javascript
import { withoutRules, addRules } from 'react-static/lib/webpack/rules';
import loadJavascript from 'react-static/lib/webpack/rules/loadJavascript';
import loadFonts from 'react-static/lib/webpack/rules/loadFonts';
import loadCSS from 'react-static/lib/webpack/rules/loadCSS';
import loadAnything from 'react-static/lib/webpack/rules/loadAnything';
```

The first two are transformers like introduced in `2.0.0.`. This concept does not differ except of the fact, that transformers does not longer need to return a valid webpack configuration object. I made the code also a bit simpler to understand.  a336e3c

Passed as webpack configuration object they can be chained together and allow us to write something like this: 
```javascript
import { withoutRules, addRules } from 'react-static/lib/webpack/rules';
import loadJavascript from 'react-static/lib/webpack/rules/loadJavascript';

export default {
    ...
    webpack: (config, args) => addRules(withoutRules(config), loadJavascript, args) //args like { stage } have to be passed as third argument to addRules() if the loader function requires them like ./rules/loadCSS.js
}
```

```javascript
import { withoutRules, addRules } from 'react-static/lib/webpack/rules';

export default {
    ...
    webpack: config => addRules(withoutRules(config), {
						loader: 'file-loader',
						test: /\.(fancyFileExtension)$/,
						query: {
							limit: 10000,
							name: 'static/[name].[hash:8].[ext]',
						},
					})
}
```

Now, it is also possible to use a transformer for debbuging purposes or just unexpected things:
```javascript
export default {
    ...
    webpack: config => console.log(config.module.rules)
};
```

Please let me know if there is anything I need to refactor or if you have further questions! :smile:


gagan0723:
As per @alexiscordova in #70:
> I think the default option should be one of the following:
> * Anything that is not supposed to be compiled (images, fonts, etc.) in src will move to dist (essentially 
    following the 2nd option @calvinrbnspiess suggested)
> * Standard image filetypes in src will move to dist, but anything else has to be a custom loader.

it's possible to merge ``` loadAnything.js ``` and ``` loadFonts.js ```.

Plus names like ``` cssLoader ```, ``` jsLoader ``` and ``` fileLoader ``` are more appropriate.

src/webpack/rules/index.js
```diff
@@ -0,0 +1,21 @@
+import loadJavascript from './loadJavascript'
+import loadCSS from './loadCSS'
+import loadAnything from './loadAnything'
```
gagan0723: You did not import and export ``` loadFonts.js ```.

calvinrbnspiess: Because every file that is not supposed to be compiled will be processed through `loadAnything.js` it would be redundant to include a separate loader doing the same thing. In case you want to modify the loader toolchain I included the `loadFonts.js' so you can swap the `loadAnything.js` with it. That would allow you to configure loaders for every file-type not handled by the provided loaders 'loadJavascript.js`, 'loadCSS.js` and `loadFonts.js`.
Maybe there will be an missing `loadFiles.js` that could handle only images, videos, etc. but I think  it wouldn't be a hazzle to write every other loaders yourself in case you want a custom toolchain.

Perhaps, it is a good idea to describe this behavior more precisely in the `README.md`

calvinrbnspiess:
> Plus names like cssLoader, jsLoader, fileLoader and fontLoader are more appropriate.
I will rename them

src/webpack/rules/index.js
gagan0723: ### Sounds great.

tannerlinsley:
This looks awesome. First things first, let's get the commit so it's not needlessly modifying all of those binary files (mostly in /media), and also not modifying the package.json spacing.

tannerlinsley:
Sorry things are changing out from underneath you so much haha.  I'm really trying to keep these bug fixes to a minimum until we can get this merged. The lib is just flexing so much right now ;)

calvinrbnspiess:
Haha yes I notice :laughing:
I was trying to resolve the issues with the touched files and picked the specific commits out - didn't thought of the master changes.

I removed the `.gitattributes` file and the commit about the media files, is this ok? Maybe it is still a good idea to insert such a  `.gitattributes` later on, so that you don't have any problems working on different systems. At beginning every file I saved in my editor was converted to CRLF

tannerlinsley:
Looks great. Can you think of any way we could support the current api and post a pre-deprecation notice? I would love to not have to increment major versions again, but if we have to, we will :) Thoughts?

calvinrbnspiess:
I did include the plugins folder, however it wouldn't be possible to use them without source-code changes. Due to the default configured loaders they are breaking. You can fix them by removing all loaders except of the `jsLoader`:
```javascript
import { withoutRules, addRules } from 'react-static/lib/webpack/rules';
import jsLoader from 'react-static/lib/webpack/rules/jsLoader';
import withCssLoader from 'react-static/lib/plugins/withCssLoader';
import withFileLoader from 'react-static/lib/plugins/withFileLoader';

export default {

        ...

	webpack: [
		withoutRules,
		config => addRules(config, jsLoader),
		withCssLoader,
		withFileLoader,
	],
```

We can hack this by removing the other loaders in the deprecated plugins, but I think it would be a better idea to undo my recent commits. So we have to increase major version again.

tannerlinsley:
That's fine. Let's polish this up and do that today

calvinrbnspiess:
Should be fine now.

tannerlinsley:
Awesome. Are the examples up to date with these changes yet?

calvinrbnspiess:
I didn't look at them yet
I'm going to test them and update if neccessary

calvinrbnspiess:
Can't test them right now but I think that will do the trick :blush:

tannerlinsley:
Looks great!

PR #77, opened 2017-10-23
[WIP] Add Intersection Observer support to Prefetch component

EmilTholin:
Thank you for your hard work on react-static. I get goosebumps just experimenting with this masterpiece of a project.

This MR adds a new component `PrefetchWhenSeen`, which will fire the prefetch of its path when it is visible in the view. This opens up for powerful techniques like e.g. prefetching a page when the user first sees a link to it:

```javascript
<PrefetchWhenSeen path={`/blog/post/${post.id}/`}>
  <Link to={`/blog/post/${post.id}/`}>{post.title}</Link>
</PrefetchWhenSeen>
```

intersection_observer_prefetch

Is this something you are interested in? The reason why I didn't put a new prop on the `Prefetch` component and add it conditionally that way, is because `Prefetch` has no wrapping element, so I could not put a `ref` on it.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -292,6 +292,54 @@ class Prefetch extends Component {
   }
 }
 
+const ioIsSupported = typeof window !== 'undefined' && 'IntersectionObserver' in window
+const handleIntersection = (element, callback) => {
+  const io = new window.IntersectionObserver(entries => {
+    entries.forEach(entry => {
+      if (element === entry.target) {
+        if (entry.isIntersecting) {
+          io.unobserve(element)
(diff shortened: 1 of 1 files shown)


tannerlinsley:
Great work! This is what I like to see :)

tannerlinsley:
Can you please open another PR and add some documentation to the README ? :)

EmilTholin:
Thanks! Absolutely. You would like the additions to the README in a separate MR?

tannerlinsley:
Yes, since I preemptively merged this one already ;)

PR #88, opened 2017-10-24
Error with .styleTags

johnrlive:
I am no expert but when I removed ".styleTags" from line 57 and then I was able to run  "react-static start"


tannerlinsley:
Fixed in the latest version :) Thanks!

PR #98, opened 2017-10-26
Add dynamic imports example

crubier:


tannerlinsley:
Looks awesome! I'm wondering if we could easily turn this into an HOC that ships with React-Static. I think I also have an idea on how to make it work even with static export :) Thoughts?

PR #100, opened 2017-10-29
Update Readme ver to 3.x.x to prevent confusion

D1no:
Readme still states `These docs are for version 1.x.x.`. Change to `3.x.x`. While I think about it, with @calvinrbnspiess changes we need to jump another major version anyhow. And master is not on npm yet? @next points to version 2 alpha.


calvinrbnspiess:
@tannerlinsley  I discovered an important bug when building react-static with custom css last night. We should fix this bug before releasing `3.0.0`.  Right now I'm trying to figure out the exact cause but it seems not easy.

PR #104, opened 2017-10-29
Add support for hashed JavaScript name

EmilTholin:
This PR changes the default name of the script from `app.js` to `app.[hash:8].js`, and would resolve https://github.com/nozzle/react-static/issues/84.

`<script async src="/app.js" />` has been removed from the `Body` component, and the resulting html from `renderToStaticMarkup` is considered a template. This template is then used by the added `HtmlWebpackPlugin` and `ScriptExtHtmlWebpackPlugin` plugins to inject a script tag with the name `app.[hash:8].js`.


tannerlinsley:
When I'm testing this, all I see loaded in the browser in dev mode is a bunch of base64. Any ideas here?

EmilTholin:
Interesting. I'm not quite sure I fully understand. Do you mean that the script is inlined as base64 somehow? Is it working at all for you?

hash_changes

tannerlinsley:
The index.html text content === 
```
data:text/html;base64,PGh0bWwgbGFuZz0iZW4tVVMiIGRhdGEtcmVhY3Ryb290PSIiPjxoZWFkPjxtZXRhIGNoYXJTZXQ9IlVURi04Ii8+PG1ldGEgbmFtZT0idmlld3BvcnQiIGNvbnRlbnQ9IndpZHRoPWRldmljZS13aWR0aCwgaW5pdGlhbC1zY2FsZT0xIi8+PC9oZWFkPjxib2R5PjxkaXYgaWQ9InJvb3QiPjwvZGl2PjwvYm9keT48L2h0bWw+<script type="text/javascript" src="/app.e4e42670.js"></script>
```

screen shot 2017-10-30 at 2 24 02 pm

EmilTholin:
Very peculiar. Are you testing this on an existing site, or using a starting example?

tannerlinsley:
A brand new `basic` example

tannerlinsley:
Wanna jump on the slack org and we can figure it out?

EmilTholin:
The basic example is broken! Remove `import logoImg from '../logo.png'` and `` from `containers/Home.js`, and `import './app.css'` from `App.js` and it works. The basic example needs a few more loaders.

tannerlinsley:
Does the html-webpack-plugin rely on an html-loader to work? I think what is happening is that the `url-loader` is somehow getting ahold of the html template, and transforming it to base64 since it's smaller than the limit.... WTF haha

EmilTholin:
What. The. Heck.
I will look into this in about 30 minutes or so!

tannerlinsley:
I think I figured this out.  It was a bad reaction between two PR's. It'll be fixed in my next release :)

EmilTholin:
Awesome. I haven't even been able to reproduce it myself yet. Looking forward to seeing the solution!

PR #109, opened 2017-10-31
Using PORT from process.env instead, if it exists

diegolameira:
I'm using hotel to get things running without port/address collision, then I need to pass a port by to get things working as expected while developing my application.

I'm an experienced angular developer in a starting react journey in the company I work for, I learned NextJS and had just found React Static hours ago as a similar (but it really sounds better) SSR framework, by the way.

Thnx, keep rocking. Best regards.

### DIFF
--- a/src/configServer.js
+++ b/src/configServer.js
@@ -7,8 +7,7 @@ import { findAvailablePort, getConfig } from './utils'
 export async function startConfigServer () {
   const config = getConfig()
   // scan a range
-  let port = await findAvailablePort(8000);
-  port = process.env.PORT || port;
+  const port = process.env.PORT || await findAvailablePort(8000)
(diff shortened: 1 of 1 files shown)


src/configServer.js
```diff
@@ -7,7 +7,8 @@ import { findAvailablePort, getConfig } from './utils'
 export async function startConfigServer () {
   const config = getConfig()
   // scan a range
-  const port = await findAvailablePort(8000)
+  let port = await findAvailablePort(8000);
+  port = process.env.PORT || port;
```
EmilTholin: You need to remove the semicolons for the linter to get happy. :)

D1no: in your working copy of the repo, you can also just run
`yarn test --fix`
to automatically fix all linting issues.

PR #112, opened 2017-10-31
FIX TypeScript Example: Babel exclude undefined

D1no:
Since the format was changed, the exclude is undefined from the rules array. Leading babel to crash by processing `node_modules` and probably the rest of the internet if we let it... 😄

### DIFF
--- a/examples/typescript/static.config.js
+++ b/examples/typescript/static.config.js
@@ -53,7 +53,7 @@ export default {
         oneOf: [
           {
             test: /\.(js|jsx|ts|tsx)$/,
-            exclude: config.module.rules[0].exclude,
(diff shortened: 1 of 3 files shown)


D1no:
The red build is not my fault ↴

```bash
/home/travis/build/nozzle/react-static/src/webpack.js
  50:33  error  'args' is defined but never used  no-unused-vars
✖ 1 problem (1 error, 0 warnings)
```

D1no:
Linting fixed with PR https://github.com/nozzle/react-static/pull/113

D1no:
@tannerlinsley green — can be merged without afterthoughts

PR #113, opened 2017-10-31
Set sourceMap for Style-Loader Instances

D1no:
Post-Css Loader only has one instance in the webpack build process if it is used multiple times. The options are only set once. To be able to use post-css to its fullest, enable sourceMaps so path-throughs work properly when other style loaders use post-css.

I.e. this is necessary to have a working less loader without dublicating the default cssLoader.


tannerlinsley:
Beautiful

PR #114, opened 2017-10-31
Example: Less with Ant Design

D1no:
> Hint: This is also a good starting point for anybody who wishes to build a sass etc. example.

Ant Design is a TypeScript and Less based [React UI Component Library](https://github.com/ant-design/ant-design) . It can be challanging to set it up with tree shaking (only require the styles that you really import) and be able to customize it (overwriting its theme variables; using styled-components to create an individual feel) without relying on css imports.

ant-design-react-static

Ant Design is written in TypeScript, which means the typescript code lens always tells you what properties and functions components accept. TypeScript is enabled to make use of that, though this example can also just be used in plain JS and/or in a mix with TypeScript.


tannerlinsley:
Let me know when this is ready to merge

examples/less-antdesign/static.config.js
```diff
@@ -0,0 +1,263 @@
...
+
+    /*
+    * In the future, this segment will change or be unnecessary.
+    * Whitelisting Extensions from pre-parser and universal url-loader
+    * */
+
+    // Tell pre-processor (babel) to no-op styles since, though they are
+    // imported as modules, are not parseable code.
+    require('babel-register')
+    require.extensions['.less'] = () => {}
```
D1no: @tannerlinsley Should be ready now! However, for the time being, we don't get around this one.

D1no:
@tannerlinsley tested — can be merged. We'll just pick this up again when changes to the pre-processor were made. We use exactly this in production.

Issue #108, opened 2017-10-30
Add example for react-apollo


BjornMelgaard:

tannerlinsley:
Anyone is free to PR this as an example :)

crubier:
Here is the PR !

https://github.com/nozzle/react-static/pull/120

PR #120, opened 2017-11-04
Add apollo redux example

### LINKED ISSUE #22: Add  example using apollo and redux
Amazing work ! A combined example with apollo and redux would be fine !

crubier:
Pretty self explanatory :-)

Fixes #22 and its duplicate #108


crubier:
This uses the new apollo client v2 which is completely independent from redux, so it is very easy to get rd of either redux or apollo if someone wants to use only one of them.

crubier:
Note that the example raises the following warning during build:

```
WARNING in ./node_modules/encoding/lib/iconv-loader.js
9:12-34 Critical dependency: the request of a dependency is an expression
 @ ./node_modules/encoding/lib/iconv-loader.js
 @ ./node_modules/encoding/lib/encoding.js
 @ ./node_modules/node-fetch/lib/body.js
 @ ./node_modules/node-fetch/index.js
 @ ./src/connectors/apollo.js
 @ ./src/App.js
 @ ./src/index.js
```

But this is only a warning, and everything works just fine as intended.

tannerlinsley:
This is awesome. Few quick questions:
- Here, you're using apollo to fetch data only at runtime, not at build time. Is that intentional?
- I would love to see this example split out into the two smaller examples as well. Redux. Then apollo.

crubier:
Yes, I only use apollo to fetch data at runtime, this is intentional. For my use case at least, creating one static page for each possible apollo query result is out of the question (that would be 100 000 static pages) so I didn't even think of it. I agree that this would be interesting though.

I will do the split tomorrow: apollo, redux, and apollo-redux

crubier:
In order to (maybe) create the apollo version with fetching at build time, I would like to know one thing which is not totally clear to me right now: the `getProps` functions defined in `static.config.js` are only called at build-time to generate the `routeData.json` files, right ? 

I believe the consequence of this would be that I would need to vastly different ways to fetch from graphql: once using `apollo-client` directly in `static.config.js` at build-time, and another using `react-apollo` (like in my commits above) at run-time.

This does not seem ideal to me, and for my use case I would rather generate a "placeholder" page in the static build so that the page feels fast to load. I am using react-static to build an hybrid website + single page application

tannerlinsley:
Makes sense. What you are describing would also be valuable. Having a
statically cached version, but also updating at runtime. In the end it's
more about what works for your needs.

crubier:
I added the redux only example and the apollo only example

crubier:
After working several hours on the example with apollo results in the static build, I can say that I did not found any clean way to do it. They all feel hacky to me and I don't think it is a good idea. It really feels like it goes against apollo, or try to go around it in some cases. Overall it leads to a lot of overhead.

If I HAD to serve rendered pages using Apollo, I would probably go with the off-the-shelf apollo SSR facilities. 

So I think the 3 examples in this PR can be merged without waiting for the apollo-in-static-build example that will not arrive soon.

tannerlinsley:
Sounds good :) thanks for trying though. You rock

tannerlinsley:
Wanna add these to the new examples section n the readme?

tannerlinsley:
Spectacular!

D1no:
Just want to add my two cents, that anybody using that apollo example would assume that static data is generated. Its react static after all - not sure what the status is now

D1no:
So i would root for at least as hackish first Implementation. Maybe from an arbitrary endpoint from https://www.graph.cool

pie6k:
Is there a way to avoid `iconv` Critical dependency error?

```
./node_modules/encoding/lib/iconv-loader.js
9:12-34 Critical dependency: the request of a dependency is an expression
```

PR #122, opened 2017-11-04
Export webpack config function

crubier:
This fixes https://github.com/nozzle/react-static/issues/101 for me. With this in place, I can now get the webpack config for my `.storybook/webpack.config.js` from the `webpackConfig` function exported from `react-static/webpack` .

### DIFF
--- a/src/webpack.js
+++ b/src/webpack.js
@@ -10,7 +10,7 @@ import { getStagedRules } from './webpack/rules'
 
 // Builds a compiler using a stage preset, then allows extension via
 // webpackConfigurator
-export async function buildCompiler ({ config, stage }) {
+export function webpackConfig ({ config, stage }) {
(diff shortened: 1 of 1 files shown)


crubier:
The second commit is here to also fix https://github.com/nozzle/react-static/issues/121 at the same time :-)

crubier:
Is it possible to publish the npm module as 4.0.2 now ?

daniellizik:
I'm using 6.0.18 and am trying to do this in a simple test file

```js
// foo.js

require('react-static/webpack');
```
but node throws 
```
Error: Cannot find module 'react-static/webpack'
```
I'm probably doing something really dumb...how do you import the react-static webpack config? I'm trying to see if I can use `react-styleguidist` with the out-of-the-box webpack config.

PR #125, opened 2017-11-05
Properly set $NODE_ENV using cross-env

SShrike:
This patch properly sets $NODE_ENV using cross-env to launch react-static with the proper environment variables set. Without this patch, some libraries and utilities, such as react-hot-loader, may leave unnecessary code behind *even when* an optimised production build has taken place.

I tried to see if I could fix this without having to push cross-env into the end user's package.json, but found no solution. There probably is a better solution though, but I'll leave that to the people who are more experienced for JavaScript/Webpack/etc. than me :p

This prevents (albeit minor) problems such as gaearon/react-hot-loader#367.

### DIFF
--- a/examples/basic/package.json
+++ b/examples/basic/package.json
@@ -4,12 +4,13 @@
   "main": "index.js",
   "license": "MIT",
   "scripts": {
-    "start": "react-static start",
-    "build": "react-static build",
+    "start": "cross-env NODE_ENV=development react-static start",
+    "build": "cross-env NODE_ENV=production react-static build",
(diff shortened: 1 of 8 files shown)


tannerlinsley:
Isn't there a way to do this in the cli in one place?

SShrike:
I tried setting the environment variable in the build command JS file, but that didn't work. I can try it again though and see if I'd just fucked it up :p

tannerlinsley:
This now being set at the script level from another PR. :)

PR #127, opened 2017-11-05
Replace regenerator with fast-async

EmilTholin:
This PR replaces `regenerator` with `fast-async`. `fast-async` just desugars `async` and `await` to Promises at compile time, rather than include a runtime. This makes the builds faster, the JavaScript bundle lighter, and most importantly the end users on mobile devices happier, because they don't get the `regenerator` runtime.

This PR also replaces the deprecated `latest` with `env`, and replaces the `stage-0` preset with appropriate plugins, to not load the `transform-async-generator-functions` plugin.

I picked the following plugins from the `stage-0` preset, going by the code base and existing examples:

- ~~transform-do-expressions~~
- ~~transform-function-bind~~
- ~~transform-class-constructor-call~~
- transform-export-extensions
- syntax-dynamic-import
- transform-class-properties
- ~~transform-decorators~~
- transform-object-rest-spread
- ~~transform-async-generator-functions~~
- ~~syntax-trailing-function-commas~~
- ~~transform-async-to-generator~~
- ~~transform-exponentiation-operator~~


tannerlinsley:
Excellent!!! Tested and working great for you?

EmilTholin:
Yep! The `dynamic-imports` example worked like a charm.

tannerlinsley:
Do you know if tree shaking works?

tannerlinsley:
If not, we may need to turn off babel's modules

tannerlinsley:
Also. Does this plane the nodent runtime in every file? Or just imported once? Does Babel-register still work after after removing Babel runtime?

EmilTholin:
I checked out `master` and built a fresh `dynamic-imports`, and did the same with this branch. The js bundle was 3 KB smaller with this branch, so it seems to be working.

`fast-async` is in `spec` mode in this PR, so there is no nodent runtime or Prototype changes.

I removed the `transform-runtime`. Is there something I overlooked here?

EmilTholin:
It seems like polyfills are not added correctly. I will take a closer look.

tannerlinsley:
I would really like to know 100% that tree shaking is happening still. Even a simple test that it can remove an npm package that is imported but not used.

EmilTholin:
I hear you. I'll put a [WIP] on this PR, I'll have to test some more. I can ping you when I feel comfortable with it.

tannerlinsley:
Sounds good. Super excited for it.

EmilTholin:
I have researched this further, and I don't think there currently is any tree shaking in `master`. 

Implementing e.g. the [cube and square example in the Webpack docs](https://webpack.js.org/guides/tree-shaking/) in a new `basic` example includes both cube and square in the minified script on `master`.

Doing the same procedure with the `regenerator_to_fast_async` branch also includes both cube and square in the minified script, but it is a few KBs smaller because there is no `regenerator` runtime.

I think the crux of it all is that `latest` might not be configured properly. I think it should be:

```javascript
[
"latest", {
  "loose": true,
  // "modules": false,  I think this is misplaced
  // This will make it so no modules are transformed, but react-static no longer works.
  "es2015": {
    "modules": false
  }
}],
```

I'm not entirely sure what the solution is yet, but [this might give some useful clues](https://github.com/babel/babel-preset-env/issues/279).

tannerlinsley:
So what happens if you set modules to false?

EmilTholin:
Then it doesn't know how to interpret `import`.

unexpected_token_import

SShrike:
I've been experimenting around with cleaning up the Babel config myself, and I can confirm that I didn't seem to find any evidence of tree shaking working either.

Also, is this *really* needed for every site? Could it make more sense to have this optional? Also, you may as well just use the stage-2 preset instead of picking the transforms manually.

tannerlinsley:
@SShrike I thought that by using any `state-n` preset, babel now warns you to use `present-env` instead. No?

Generally, I would absolutely say that tree shaking is recommended for every single site. Marketing sites more than SPA's need to be very size-aware and as small as possible. Tree shaking would ensure that.

EmilTholin:
The `stage-2` preset includes every plugin from `stage-3` as well, which contains `transform-async-generator-functions` and `transform-async-to-generator` which would not work with `fast-async`.

SShrike:
@tannerlinsley Nah, you still have to use the stage presets if you want any features that aren't yet in the ES spec.

@EmilTholin Ah yes, good point. I'm still not entirely convinced on this being a *default* though, seems kind of like a case of premature optimisation that not everyone needs as well as breaking the rule of least surprise because now someone might look in the config and say "what is this fast-async thing?", "why are there hand picked presets?", etc.

tannerlinsley:
In all honesty, I don't really have any preference how we achieve these things, but these are the things I want to happen:

- Tree Shaking
- No more babel warnings
- A generous, ~90% use case babel config that will cover a majority of the communities' expectations like async/await, spread/rest, es7 class transforms, etc.

To be clear, I think the fast-async thing is cool, but I agree, it's not a necessity. I don't know much about it to be honest. If there is already doubt or concern being raised that it could become a default, then I want to pump the brakes a bit.

Thoughts?

SShrike:
Well what I was working on was essentially the following:

- Remove Babel warnings.
- Use preset-env with the same settings as Autoprefixer.
- Remove the unneeded presets for just stage-3 and env.
- Get tree shaking working.
- Upgrade to the Babel 7 beta (not *required*, just was testing it and the JS configs).

tannerlinsley:
Love that plan, including upgrading to Babel 7 beta.

SShrike:
Thanks! There's one blocker though, gaeron/react-hot-loader#693.

tannerlinsley:
Sounds good. Can't wait haha. So @EmilTholin, I think from here it would be better suited to work with @SShrike on implementing the above list. Let's hold off on fast-async.

SShrike:
Ok, provided that PR I linked gets merged, I should be able to get a WIP PR out by Monday next week. Although I'm also busy with school and other things, so don

EmilTholin:
Sounds good. When that PR gets merged, that sounds like a great plan going forward.

I don't share your skepticism towards `fast-async` though. [Neutrino uses it](https://github.com/mozilla-neutrino/neutrino-dev/blob/master/packages/neutrino-preset-web/index.js#L39), Gatsby will add is soon, and even [the Zen masters of Babel are looking into it](https://twitter.com/left_pad/status/924487010575880192). But I get your point.
 
I'm working on a new example for the `examples` folder. Maybe I should use `fast-async` there instead, to show that it is fully customizable in `react-static`, and close this MR?

EmilTholin:
I'll open up a new PR if I have some tree shaking insight!

PR #134, opened 2017-11-07
Babel 7, add tree shaking, replace latest with env

EmilTholin:
This PR adds tree shaking and replaces the deprecated babel preset `latest` with `env`.

There is a slight misconfiguration of the `latest` preset:

```javascript
["latest", {
  "loose": true,
  // "modules": false // This shall be configured in "es2015" below
  // This will make it so no modules are transformed.
  "es2015": {
    "modules": false
  }
}]
```

Not transforming the modules to `commonjs` sets the code up nicely to be tree shaken, but the code still has to be transformed in order to run in Node. I therefore moved the `.babelrc` to a custom preset `.babelrc.js`, so `modules` could be conditionally turned on and off depending on if the code shall be used by Node or not.

Babel works by taking the closest `.babelrc` of the file it tries to import, so I turned of this feature with `babelrc: false` as an argument to `babel-register`, so that `static.config.js` is not loaded with the user's own babel config.

Webpack was now properly tagging parts that could be shaken off (e.g. `/* unused harmony export square */`), but they were still present in the resulting minified bundle. It turns out that `react-hot-loader/babel` makes tree shaking impossible, so removing that plugin for production made tree shaking finally work.

There are some exciting additional features in the `env` preset for babel 7, but I think it's best to wait with that until [react-hot-loader is up and running for babel 7](https://github.com/gaearon/react-hot-loader/pull/693).


tannerlinsley:
This looks great. Are there any negative side effects from using preset instead of extend in project Babel rc?

EmilTholin:
Thanks!
It was not possible to have `react-static` use `isNode` and still extend it in the examples. This works just as good. Do you know of any other nice way to achieve it?

SShrike:
Shouldn't this be using `stage-2`, not `stage-0` as per the previous discussion?

EmilTholin:
`stage-0` will only add the additional options to use `transform-do-expressions`,
`transform-function-bind` and `transform-export-extensions`, and will not clutter the bundle is they are not used, so I don't see an issue with it. I can change to `stage-2` if you like.

SShrike:
That seems reasonable enough, I'll cede to you for the moment 😉.

.babelrc
```diff
@@ -1,38 +1,5 @@
...
-        "stage-0",
-        "stage-3",
-        "react"
-      ]
-    }
-  }
-}
+    ["./.babelrc.js", { "isNode": true }]
+  ]
+}
```
SShrike: World's smallest nitpick: You're missing the final newline on this file.

SShrike:
Also I never said this yet, but great work!

.babelrc
EmilTholin: Ah, well spotted.

EmilTholin:
Thanks!

tannerlinsley:
@EmilTholin, it does work well, I was hoping not to perform any breaking changes for this upgrade though ;)

EmilTholin:
Totally understand. I tried to have `.babelrc` as just `{ "presets": ["./.babelrc.js"] }`, and add `--presets=./.babelrc.js` to `build` and `watch` in order to not change the examples, but that didn't quite work sadly. I can give it some more thought if you like.

tannerlinsley:
If it's a breaking change that's okay. Just want to try and avoid it if
possible. Let me know if it's not doable?

EmilTholin:
I understand. I can research it a bit more today after work. I will keep you updated on how it goes.

EmilTholin:
I have researched this some more, and I think I've stumbled upon something awesome.

Instead of conditionally transforming modules to `commonjs` with `isNode`, we can throw `commonjs` out of the window entirely by using [@std/esm](https://github.com/standard-things/esm). This way both environments can use ES6 and the same `.babelrc`, so no breaking change will be introduced.

This also means that `react-static` itself will be shaken, because it is no longer `commonjs` modules that has to be included in its entirety.

tannerlinsley:
Outstanding!!!

EmilTholin:
I still need to test this more thoroughly. Shall I ping you when it feels solid?

tannerlinsley:
Yeah, as soon as it's ready, we can put it through the ringer with some experimentation and existing sites.

tannerlinsley:
@EmilTholin Any updates on this? I would be lying if I said I wasn't super stoked for this PR.

EmilTholin:
It's going great! The last missing piece of the puzzle with [`@std/esm` and `@babel/register`](https://github.com/standard-things/esm/issues/143) is soon complete, and then Babel 7 will come with this PR as well.

tannerlinsley:
Awesome. Can't wait!

SShrike:
Have to say I'm quite impressed so far, probably a better solution than I would've ended up having done :p

SShrike:
Babel v7.0.0-beta.32 is out which includes jdalton's patch to babel-register, so everything should be all set at this point right?

EmilTholin:
@SShrike Yeah, the patch to [Pirates](https://github.com/ariporad/pirates) needed to be released, and then [esm](https://github.com/standard-things/esm) as well. `esm` got a new release yesterday, so we should be set! I will add it after work today.

EmilTholin:
This PR now also includes Babel 7.

tannerlinsley:
So can you confirm this is working for existing projects and fresh installs? (minus the old redbox behavior)

EmilTholin:
Yep! I've tested thoroughly with `basic` and `dynamic-imports` and fresh installs. It would of course be great if you had the time to checkout the PR and play around with it as well.

tannerlinsley:
Can do

tannerlinsley:
I'm getting an error:
screen shot 2017-11-13 at 2 00 13 pm
Thoughts?

tannerlinsley:
Ah. Looks like this needed to happen:
```javascript
// require.resolve('react-dev-utils/webpackHotDevClient'),
`${require.resolve('webpack-dev-server/client')}?/`,
```

EmilTholin:
Ah, shoot. How did you get this error? I can't reproduce it. [This might give some clues](https://github.com/facebookincubator/create-react-app/issues/3332).

PR #135, opened 2017-11-08
Enable customization of devServer config

rywils21:
- Makes development server port configurable via command line
- Makes development server configurable via a `devServer` object in `static.config.js`


tannerlinsley:
Awesome work on this man! I added a few more enhancements to show if the desired or default port ends up being unusable, and I also added a way for the environment variable `HOST` to be used just like `PORT`. 

Awesome work, again. You rock man. You're gonna be a great OSS contributor.

PR #136, opened 2017-11-08
Allow path customization

tannerlinsley:


src/utils.js
```diff
@@ -32,24 +31,59 @@ export const findAvailablePort = start =>
...
+  // Resolve all paths
+  const paths = {
+    ROOT,
+    LOCAL_NODE_MODULES: path.resolve(__dirname, '../node_modules'),
+    SRC: resolvePath(config.paths.src),
+    DIST: resolvePath(config.paths.dist),
+    PUBLIC: resolvePath(config.paths.public),
+    NODE_MODULES: resolvePath('node_modules'),
+    PACKAGE: resolvePath('package.json'),
+    HTML_TEMPLATE: path.join(resolvePath('dist'), 'index.html'),
```
EmilTholin: Really cool stuff!
Maybe this should be `HTML_TEMPLATE: path.join(resolvePath(config.paths.dist), 'index.html')`, or am I misunderstanding something?

tannerlinsley: You're right, it should :(

PR #137, opened 2017-11-08
Fix webpack production config destructuring

EmilTholin:
`react-static start` is working great after the path refactor, but `react-static build` crashed because the `config` is not passed properly to the rules function.


tannerlinsley:
:)

Issue #138, opened 2017-11-08
Error in build if 404 route not provided


mplis:
If an `is404` route is not in the routes array returned from `getRoutes`, running the build command results in this error:
```
Warning: React.createElement: type is invalid -- expected a string (for built-in components) or a class/function (for composite components) but got: boolean.
    in Route
    in Routes
    in div
    in div
    in Router
    in StaticRouter
    in Router
    in _default
    in InitialPropsContext
    in ContextualComp
TypeError: element.type.toLowerCase is not a function
    at ReactDOMServerRenderer.renderDOM (/Users/mplis/personal/misc/my-static-site/node_modules/react-dom/cjs/react-dom-server.node.development.js:2776:28)
    at ReactDOMServerRenderer.render (/Users/mplis/personal/misc/my-static-site/node_modules/react-dom/cjs/react-dom-server.node.development.js:2755:23)
    at ReactDOMServerRenderer.read (/Users/mplis/personal/misc/my-static-site/node_modules/react-dom/cjs/react-dom-server.node.development.js:2722:19)
    at renderToString (/Users/mplis/personal/misc/my-static-site/node_modules/react-dom/cjs/react-dom-server.node.development.js:2980:25)
    at Object.renderToHtml (/Users/mplis/personal/misc/my-static-site/node_modules/react-static/lib/utils.js:107:14)
    at _callee$ (/Users/mplis/personal/misc/my-static-site/node_modules/react-static/lib/static.js:236:39)
    at tryCatch (/Users/mplis/personal/misc/my-static-site/node_modules/regenerator-runtime/runtime.js:65:40)
    at Generator.invoke [as _invoke] (/Users/mplis/personal/misc/my-static-site/node_modules/regenerator-runtime/runtime.js:299:22)
    at Generator.prototype.(anonymous function) [as next] (/Users/mplis/personal/misc/my-static-site/node_modules/regenerator-runtime/runtime.js:117:21)
    at step (/Users/mplis/personal/misc/my-static-site/node_modules/react-static/lib/static.js:120:191)
    at /Users/mplis/personal/misc/my-static-site/node_modules/react-static/lib/static.js:120:437
    at Promise (<anonymous>)
    at /Users/mplis/personal/misc/my-static-site/node_modules/react-static/lib/static.js:120:99
    at /Users/mplis/personal/misc/my-static-site/node_modules/react-static/lib/static.js:356:30
    at Array.map (<anonymous>)
[...]

PR #139, opened 2017-11-08
Ensure Template is a valid React element before trying to render it

mplis:
### Is this a bug report?

This bug is detailed in #138. This is the simplest fix I could think of for this bug. It ensures that `Template` is a valid React element before trying to render it and returns `undefined` otherwise.
 
 ### Environment
 
```
-> % react-static -v && node -v && npm -v
4.0.1
v8.3.0
5.3.0
```
 
Operating system: macOS Sierra 10.12.6
 
 ### Steps to Reproduce
 
1. Run `react-static create`
2. Choose `basic` template (it should be a bug with any template, but I haven't confirmed)
3. Remove this from `static.config.js`:
```
{
    is404: true,
    component: 'src/containers/404',
},
```
4. Run `npm run build`
 
 ### Expected Behavior

I expected the `build` command to run without errors
 
 ### Actual Behavior
 
The `build` command threw an error

### DIFF
--- a/src/static.js
+++ b/src/static.js
@@ -281,7 +281,7 @@ export const prepareRoutes = async config => {
       if (!Template) {
         Template = getTemplateForPath('404')
       }
-      return <Template {...props} />
+      return Template && <Template {...props} />
     }} />
   `
 


tannerlinsley:
Looks simple enough to me :) Good work!

PR #140, opened 2017-11-08
Typings for React-Static for full TypeScript Support

D1no:
Bundles full TypeScript support. Since React-Static is just a thin layer around existing well typed libraries, we are relaying the type information accordingly (react-router, react-helmet).

This enables TypeScript / WebStorm users to use autocomplete and propper reference to the react-static library.

Since this is a declaration file that doesn't engage with any implementations (just a declaration), this does not effect code in anyway.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,7 @@
   "name": "react-static",
   "version": "4.0.1",
   "main": "lib/index.js",
+  "types": "src/index.d.ts",
   "license": "MIT",
   "bin": {
     "react-static": "./bin/react-static"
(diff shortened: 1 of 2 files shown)


D1no:
No Idea what that build error means:
> The command "eval yarn " failed. Retrying, 2 of 3.

tannerlinsley:
I have never seen that error.  Ever. I have no idea what it is.

tannerlinsley:
Regardless, this looks great!

D1no:
Mh — obviously its impossible for that PR to touch anything.... so what to do? Can you merge it? :) Than I could throw this out of our internal system where we used this declaration file for a while now :)

tannerlinsley:
Merged!

PR #161, opened 2017-11-13
[WIP] Babel 7, tree shaking, replace latest with env

EmilTholin:
_[There is a slight regression in Babel 7](https://github.com/facebook/regenerator/pull/319) that needs to be fixed before this PR can be merged._ 

This PR moves to Babel 7, moves away from `commonjs` to ES6 with `esm` to allow for tree shaking, and replaces the deprecated `latest` preset with the `env` preset.

### DIFF
--- a/src/static.js
+++ b/src/static.js
@@ -17,10 +17,6 @@ import { DefaultDocument } from './RootComponents'
 export const exportRoutes = async ({ config, clientStats, cliArguments }) => {
   // Use the node version of the app created with webpack
   const Comp = require(glob.sync(path.resolve(config.paths.DIST, 'static.*.js'))[0]).default
-  const manifestScript = glob
-    .sync(path.resolve(config.paths.DIST, 'manifest.*.json'))[0]
-    .split('/')
-    .pop()
 
   const DocumentTemplate = config.Document || DefaultDocument
 
@@ -209,9 +205,6 @@ export const exportRoutes = async ({ config, clientStats, cliArguments }) => {
             {showHelmetTitle && head.title}
             {head.meta}
             {head.link}
-            {manifestScript && (
-              <link rel="manifest" href={`${config.publicPath}${manifestScript}`} />
-            )}
             {process.env.extractedCSSpath && (
               <link rel="stylesheet" href={`${config.publicPath}${process.env.extractedCSSpath}`} />
             )}


babel-preset.js
```diff
@@ -0,0 +1,36 @@
...
+    }]
+  ]
+
+  if (process.env.NODE_ENV === 'development') {
+    plugins.unshift(r('react-hot-loader/babel'))
+  }
+
+  return {
+    presets: [
+      [r('@babel/preset-env'), {
```
SShrike: Maybe re-use the browser list that Autoprefixer uses?

SShrike: image

talking about this specifically

SShrike: image

Could just extract the list from `src/webpack/rules/cssLoader.js` and then use it in both the Babel config and CSS loader config, just to reduce repetition.

EmilTholin: Thanks for the review @SShrike!
Yeah, I alludes to that in the description, "... super cool feature of `useBuiltIns: 'usage'` in Babel 7." The polyfill part of `target.browsers` is not really working properly for this use case yet, and since ["Without any configuration options, `@babel/preset-env` behaves exactly the same as `@babel/preset-latest`"](https://github.com/babel/babel/tree/master/experimental/babel-preset-env), I thought we could leave it as it is for now.

tannerlinsley: More along those lines, @derekperkins and I have been talking about a plugin that could determine these types of parameters based on google analytics device/browser usage thresholds. Something like, "Enter you GA key, and we'll handle the rest".

EmilTholin: That sounds really awesome!

Look what else just got opened! https://github.com/MatAtBread/fast-async/issues/46

tannerlinsley:
What's the latest on this?

EmilTholin:
Regenerator has been changed [to not have a set version of `babel-types` as a dependency](https://github.com/facebook/regenerator/pull/329), so the regression should not be a problem any more. I will have time to experiment with it on Monday.

tannerlinsley:
What's the status on this?

EmilTholin:
Haven't had the time to test is thoroughly, but it should only be a rebase and bump to `beta-36` on the Babel packages left. I will find the time in the coming week.

EmilTholin:
I did a rebase, and it got closed automatically. That's weird. Must have done something silly. I will open up a new MR.

PR #164, opened 2017-11-14
move configServer to webpack.devServer.before() and only expose one port

rileylnapier:
Hi, this is a PR to remove the `configServer` and focus on only have one server.
I am routing all `configServer` to use a prefix of __config__ instead of having a separate server.

This will allow me to host a dev server of my react-static app to my client on aws/heroku ect... and will help automate lots of fun static site generation :)

Let me know if there is a better way to handle this... or if the configServer was required for a use case that I am unaware of?


tannerlinsley:
This looks great! Have you tested with examples?

rileylnapier:
i just tested with my local app i'm building...  

also, I need to do a quick push to fix "findAvailablePort" ~ the env variable of PORT was coming in as a string and we need to cast it to number.  im testing this right now

rileylnapier:
ok, i fixed the port issue.

how should i test the examples?  is there a command i should run?  i didn't see anything about "how to dev"

tannerlinsley:
Just install a few of them fresh. Test dev. Test a build. You can use the yarn start, build, and serve scripts that come with each example.

tannerlinsley:
This is working great! Just needed to fix a few more prefix and we're good to go! Expect a release soon.

rileylnapier:
awesome!!! thanks!!

PR #175, opened 2017-11-15
Add .md & .yaml to Extension whitelist

D1no:
Extensions not getting through to webpack as they are blocked by the Babel preprocessor.
 
 ### Is this a bug report?
 
Yes 

See: https://github.com/nozzle/react-static/issues/171

### DIFF
--- a/src/commands/index.js
+++ b/src/commands/index.js
@@ -21,6 +21,8 @@ const ignoredExtensions = [
   'ogg',
   'mp3',
   'wav',
+  'md',
+  'yaml'
 ]
 ignoredExtensions.forEach(ext => {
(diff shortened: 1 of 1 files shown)


EmilTholin:
It is already [working fine](https://github.com/EmilTholin/react-static-markdown-example), or am I misunderstanding something?

D1no:
It can work depending on some individual systems (don’t ask me why). But excluding the handling of files (their extensions) that turn up through module imports is a save why to guarantee that Babel is not chocking on those non interpretable files. This is a common issue in unit tests.

The unrecognized token error from the stated issue is a prime example of this problem.

Issue #178, opened 2017-11-16
Closing script tag in prop string breaking HTML parser in browser


BeheadedKamikaze:
I have blog post content coming from Wordpress and it contains Javascript tags. This is fine and renders properly using `<article dangerouslySetInnerHTML={{ __html: post.body }} />`.

The issue is that the props get serialised into the bottom of the document, but my `post.body` string contains the fragment `</script>`, which causes the browser to stop parsing the document as a script, and the rest of the string gets parsed as HTML; the result is not pretty.

One solution would be in the `window.__routeData` to encode the slash using a backslash like `<\/script>`, which I have done manually and verified that it works. Another solution would be to break the string, like `"...<" + "/script>..."`.

Is there any way for the encoder to do this automatically?

There is nothing "special" about the `/` character so it doesn't get encoded by default - it's just rather unfortunate that when it appears between `<` and `script>`, that it breaks the context in a browser parser.

tannerlinsley:
Yeah we should probably escape stuff like that.

PR #179, opened 2017-11-16
Escape script tags in routeData (#178)

BeheadedKamikaze:
### Is this a bug report?
 
Yes

See #178 

Any version
 
 ### Steps to Reproduce
  
 1. Create a route prop with a string containing `<script></script>`
 2. Build the static site
 3. Open the static page in any browser
 
 
 ### Expected Behavior
 
The page would render correctly
 
 
 ### Actual Behavior
 
 The browser stops parsing the route props generated by react-static as soon as it encounters the character sequence `</script>`, causing the rest of the script to incorrectly be interpreted as document markup.

### The solution

Break the generated JSON string and rejoin with `+` operator to avoid the browser parsing it as a closing script tag

### DIFF
--- a/src/commands/index.js
+++ b/src/commands/index.js
@@ -22,7 +22,7 @@ const ignoredExtensions = [
   'mp3',
   'wav',
   'md',
-  'yaml'
+  'yaml',
 ]
 ignoredExtensions.forEach(ext => {
   require.extensions[`.${ext}`] = () => {}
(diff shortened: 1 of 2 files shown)


BeheadedKamikaze:
`JSON.stringify()` serialises props to a string surrounded by double quotes, so this regex searches for opening or closing script tags `<script` and `</script` and replaces them with `<"+"script` and `<"+"/script` to break and rejoin the string, so that the browser won't parse it.

tannerlinsley:
Nice job!

PR #183, opened 2017-11-18
Set NODE_ENV on start/build, use distPath for html template

EmilTholin:
### Is this a bug report?
 
Yes
 
 ### Environment
 
 1. `react-static -v`: 4.3.0
 2. `node -v`: 8.9.0
 3. `yarn --version`: 1.3.2
 
 ### Steps to Reproduce
 
 1. Add `paths: { devDist: 'local' }` to `static.config.js`
 2. Start the dev server with `yarn start`
 
 ### Expected Behavior

Creates a `local` folder and uses that instead of `dist` 
 
 ### Actual Behavior
 
Uses the `dist` folder
 
### Probable explanation
`process.env.NODE_ENV` is not set properly, so the `process.env.NODE_ENV === 'development'` check when giving `paths.DIST` a value is not working as intended. Explicitly giving it a value on `start` and `build` works great.

The `HTML_TEMPLATE` path also had to be made dynamic, and all the files were now created in the correct folder.

### Closes
https://github.com/nozzle/react-static/issues/181

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -52,18 +52,18 @@ export const getConfig = () => {
   const resolvePath = relativePath => path.resolve(path.join(ROOT, relativePath))
 
   // Resolve all paths
+  const distPath = process.env.NODE_ENV === 'development'
+    ? resolvePath(config.paths.devDist || config.paths.dist)
+    : resolvePath(config.paths.dist)
   const paths = {
     ROOT,
     LOCAL_NODE_MODULES: path.resolve(__dirname, '../node_modules'),
     SRC: resolvePath(config.paths.src),
-    DIST:
-      process.env.NODE_ENV === 'development'
-        ? resolvePath(config.paths.devDist || config.paths.dist)
-        : resolvePath(config.paths.dist),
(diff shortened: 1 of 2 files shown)


EmilTholin:
Sorry, I'm not quite sure I understand. With the value `HTML_TEMPLATE: path.join(resolvePath(config.paths.dist), 'index.html')`, two folders are created.

two_folders

Was that intended?

src/utils.js
```diff
@@ -52,18 +52,18 @@ export const getConfig = () => {
   const resolvePath = relativePath => path.resolve(path.join(ROOT, relativePath))
 
   // Resolve all paths
+  const distPath = process.env.NODE_ENV === 'development'
```
tannerlinsley: HTML template should use this distPath.

tannerlinsley:
Oh never mind. It looks like it's good to go. Just an early morning ;)

src/utils.js
EmilTholin: Yep! Totally agree. I changed to that. :)

PR #186, opened 2017-11-19
Update README.md

EmilTholin:
- Add the Cordova example to the list of examples
- Add `devServer` to the `static.config.js` overview
- Show that `renderToHtml` can be asynchronous
- Import from `react-static` instead of `react-router` where appropriate
- Fix a few typos


tannerlinsley:
Nicely done!

PR #191, opened 2017-11-22
Fix path for imports on Windows

karland:
### Is this a bug report?
 
Fixes https://github.com/nozzle/react-static/issues/184

 ### Environment
 
 
 
 1. `react-static -v`: 4.3.2
 2. `node -v`: 9.2.0
 3. `yarn --version or npm -v`: yarn 1.3.2
 
 Then, specify:
 
 1. Operating system: Windows 10
 2. Browser and version (if relevant):
 
 
 ### Steps to Reproduce
 
 
 
 (Write your steps here:)
 
 1. Run example on Windows (I used the styled-components one)
 
 ### Expected Behavior
 
Run without error. 

### Cause of issue

[path.resolve(config.paths.ROOT, template),](https://github.com/nozzle/react-static/blob/da68e23609742a38cf22b92eef07d343f2af44be/src/static.js#L241) is returning OS specific paths. On Windows they contain backslashes instead of slashes. To fix this, all backslashes need to be replaced with slashes, because even on Windows the import paths have forward slashes.

`

### DIFF
--- a/src/static.js
+++ b/src/static.js
@@ -242,6 +242,7 @@ export const prepareRoutes = async config => {
         )}'`,
     )
     .join('\n')
+    .replace(/\\/g, '/')
 
   const templateMap = `const templateMap = {
     ${templates


tannerlinsley:
Nice work!

karland:
Thanks. Would you consider to make a 4.3.3 soon, so that we are back to a working version on Windows? That would be rather great....

tannerlinsley:
I'll make a release in the next few hours :)

Issue #196, opened 2017-11-25
Add example where Progressive Web App is enabled by default


muloka:
Progressive Web Apps are faster and more reliable than traditional web pages, and provide an engaging mobile experience

It would be good to have at least one example of how to do this in react-static.

tannerlinsley:
Linked with #197

PR #197, opened 2017-11-25
Added example where Progressive Web App is enabled by default

muloka:
Using the glamorous example as a base, created a pwa-glamorous example where Progressive Web App features are enabled by default.

See issue #196


tannerlinsley:
So what's next here?  What's the state of this PR?

tannerlinsley:
Going to close this for now.  Would love to see this work out though!

danielemesh:
Anything new here?
There is no PR or a PWA example...

tannerlinsley:
I would love to see activity here too.  Might be easier after v5 lands...

danielemesh:
@tannerlinsley what is planned for v5?

tannerlinsley:
Take a look at the changelog here
https://github.com/nozzle/react-static/blob/master/CHANGELOG.md

spences10:
Is this useable?

I'll give it a try and let you know how I get on...

spences10:
So I tried this and like with the other examples I have tried I get a `ReferenceError: window is not defined` which I think is a `node` thing but I don't know what the alternative is, think it's global 🤔

PR #203, opened 2017-11-27
Added class example for route props to readme

Pixelatex:
### Is this a bug report?
 
No
 
### Why is this needed?

I noticed a few people were looking for this and the only answer is somewhere in the hour-long youtube video. Seems like something useful to add.


foodaka:
This should be added to the readme section to clarify for fellow developers :)

PR #212, opened 2017-12-02
Add .esm-cache to the .npmignore

ChrisNLott:
.esm-cache is a huge folder (~25mb) which results in the react-static package being ~35mb.

Is this folder necessary for the published package?

It is already being ignored via [.gitignore](https://github.com/nozzle/react-static/blob/master/.gitignore#L20). 

 ### Environment
 
 1. `react-static -v`: 4.3.4
 2. `node -v`: 8.9.1
 3. `yarn --version or npm -v`: yarn 1.3.2
 4. Operating system: macOS
 
 ### Actual Behavior
 
image

https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -14,6 +14,8 @@
 .env.development.local
 .env.test.local
 .env.production.local
+.esm-cache
+
 *.log
 npm-debug.log*
 yarn-debug.log*


tannerlinsley:
Awesome work!

PR #214, opened 2017-12-02
Use ReactDOM.render instead of ReactDOM.hydrate in dev

EmilTholin:
This PR changes `ReactDOM.hydrate` to `ReactDOM.render` in dev mode in all the examples, so the user doesn't get `Warning: Expected server HTML to contain a matching <div> in <div>.`, as outlined by @mattyfresh in https://github.com/nozzle/react-static/issues/144.


tannerlinsley:
Great work

PR #215, opened 2017-12-03
graphql-request example

hmeissner:
Static blog example using `graphql-request` and GraphCMS as data source.

### DIFF
--- a/examples/graphql-request/.babelrc
+++ b/examples/graphql-request/.babelrc
@@ -1,0 +1,3 @@
+{
+  "extends": "react-static/.babelrc"
+}
--- a/examples/graphql-request/.eslintrc.js
+++ b/examples/graphql-request/.eslintrc.js
@@ -1,0 +1,3 @@
+module.exports = {
+  extends: 'react-tools'
+}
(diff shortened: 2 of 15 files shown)


tannerlinsley:
Fantastic! Let me know when you're ready to merge

hmeissner:
Let me just adjust the styling to pass Travis and we should be good to go!

tannerlinsley:
Excellent!

hmeissner:
`eslint` throws a warning about https://github.com/hmeissner/react-static/blob/master/examples/graphql-request/src/pages/Post.js#L13 with `no-danger`. The `dangerouslySetInnerHTML` is there to parse the content of GraphCMS`s markdown field which is a string with html tags.

To avoid the warning I can throw in an eslint ignore along with a proper explanation of why the `dangerouslySetInnerHTML` is there. Something like this:
```
{/*
  dangerouslySetInnerHTML is here to parse a markdown field from GraphCMS.
  The field contains a string with HTML tags as GraphCMS already does the markdown parsing.
  If you want to avoid dangerouslySetInnerHTML in your project you can use any html/markdown
  parsing plugin for React like https://github.com/rexxars/react-markdown
*/}
<div dangerouslySetInnerHTML={{ __html: post.content }} /> {/* eslint react/no-danger: 0 */}
```

tannerlinsley:
That's more than okay :)

PR #217, opened 2017-12-04
revert dist path changes

poorpaddy:
Reverted the dist path changes. The old method prior to 4.3.1 worked fine just had some annoyances when utilizing https on port 443 but was more usable than forcing node environment.

### DIFF
--- a/src/commands/index.js
+++ b/src/commands/index.js
@@ -37,12 +37,10 @@ export default function () {
   }
 
   if (cmd === 'start') {
-    process.env.NODE_ENV = 'development'
     return require('./start').default()
   }
 
   if (cmd === 'build') {
-    process.env.NODE_ENV = 'production'
     return require('./build').default()
   }
 
--- a/src/utils.js
+++ b/src/utils.js
@@ -52,18 +52,18 @@ export const getConfig = () => {
   const resolvePath = relativePath => path.resolve(path.join(ROOT, relativePath))
 
   // Resolve all paths
-  const distPath = process.env.NODE_ENV === 'development'
-    ? resolvePath(config.paths.devDist || config.paths.dist)
-    : resolvePath(config.paths.dist)
   const paths = {
     ROOT,
     LOCAL_NODE_MODULES: path.resolve(__dirname, '../node_modules'),
     SRC: resolvePath(config.paths.src),
-    DIST: distPath,
+    DIST:
+      process.env.NODE_ENV === 'development'
+        ? resolvePath(config.paths.devDist || config.paths.dist)
+        : resolvePath(config.paths.dist),
     PUBLIC: resolvePath(config.paths.public),
     NODE_MODULES: resolvePath('node_modules'),
     PACKAGE: resolvePath('package.json'),
-    HTML_TEMPLATE: path.join(distPath, 'index.html'),
+    HTML_TEMPLATE: path.join(resolvePath(config.paths.dist), 'index.html'),
   }
 
   const siteRoot = config.siteRoot ? config.siteRoot.replace(/\/{0,}$/g, '') : null


tannerlinsley:
This doesn't affect react bundling? I thought it was very important that the node environment be set appropriately for react to minify and bundle correctly

poorpaddy:
Not really. React does do optimizations based on production explicit node env however in my tests it only resulted in a handful of bytes. I think it would be best that if you want to use `production` you can for most optimal output. However at least you will have the option to not only use something different but additional environments.

EmilTholin:
Without explicitly setting `NODE_ENV`, the template file will always end up in `dist`, and not `devDist`. We need another way or distinguishing between `start` and `build` in the code.

poorpaddy:
I'd rather have options more than have devDist. One is a nice to have to other provides max flexibility.

tannerlinsley:
Right now they're are plenty of code paths that rely on strict development
and production node environments though. How would we get around that?

tannerlinsley:
Annually I think I have an idea. We can make our own variables much like
the stages used in the webpack configs. Then set up some environment
fallbacks for the start and and build scripts while still allowing people
to override with tools like cross env

tannerlinsley:
Yep, that seems fair. I'll work on a separate PR. Care if I close this one?

poorpaddy:
I agree with both of you.

tannerlinsley:
K, I just committed some changes that should allow for any `NODE_ENV` to be used at any time. If no `NODE_ENV` is set, it will default to `development` on the `start` command and `production` on the `build` command.  This way, 99% of people will never have to worry about setting it themselves, but if you need to, you can easily use something like `cross-env` to handle multiple build environments.  Internally, RS now uses `REACT_STATIC_ENV` which is set to either `development` or `production` depending on the command used (`start` and `build` respectively).

EmilTholin:
Awesome!

This is how @poorpaddy uses `NODE_ENV`, taken from https://github.com/nozzle/react-static/issues/190:

```
"dev": "NODE_ENV=dev react-static build",
"cqa": "NODE_ENV=cqa react-static build",
"prod": "NODE_ENV=prod react-static build"
```

Wouldn't you still want to give [`NODE_ENV: 'production'` to Webpack](https://github.com/nozzle/react-static/commit/0561344f994e54c55f121e1bfe3a2fbaad9921dd#diff-87678bf520605f3c1638dc83c2ff66f6L33) for `build`, and `NODE_ENV: 'development'` for `start`? This way you are forced to use `NODE_ENV=production` if you don't want a massive JS bundle. Maybe I'm overthinking it?

poorpaddy:
`"start": "NODE_ENV=dev react-static start",`. Even thnough I used a different name for development it still resolves as `development`. This is fine.

The reason for use node_env rather than an option is for builds/deployment automation with Jenkins and some other tools that are essentially just running the scripts.

Here's all the scripts I'm using:
```
"lint": "eslint --format node_modules/eslint-friendly-formatter --ext .jsx --ext .js src && stylelint ./src/**/*.scss",
    "clean": "rm -rf ./dist ./coverage ./log",
    "test": "NODE_ENV=test karma start ./karma/karma.conf",
    "functional": "wdio",
    "functional:local": "node functional.js",
    "prepush": "yarn lint && yarn test",
    "start": "NODE_ENV=dev react-static start",
    "serve": "serve dist -p 3000",
    "dev": "NODE_ENV=dev react-static build",
    "cqa": "NODE_ENV=cqa react-static build",
    "prod": "NODE_ENV=prod react-static build",
    "build": "rimraf dist && yarn dev && yarn cqa && yarn prod",
    "build-preprod": "yarn dev && yarn cqa",
    "build-release": "yarn build-preprod && yarn prod",
    "release": "yarn build-release"
```

EmilTholin:
Cool. So the fact that all the `process.env.NODE_ENV === 'production'` parts of e.g. React will not be removed with your `"cqa": "NODE_ENV=cqa react-static build"` command is not really a problem then?

poorpaddy:
Was looking over the commit. Really liking what I've seen. Has been tested or should I download and give it a whirl?

EmilTholin:
I'm sure it will be released in the next minor release any time soon, but it would be great if you could try the `master` branch before that.

poorpaddy:
ok, I'm checking out both scenarios of custom vs standard now.

poorpaddy:
Seems to be working in both scenarios. i am having some unrelated issues with the ExtractTextPlugin.

tannerlinsley:
Can you fire off another issue for that?

poorpaddy:
@tannerlinsley yeah. Im still troubleshooting it. Im not convinced its an issue witg react-static yet.

poorpaddy:
@tannerlinsley confirmed my issue had nothing to do with react-static. Your changes made everything work perfectly. Ship it!

PR #218, opened 2017-12-05
add glamorous-tailwind example

deden:
Add Tailwind CSS example with CSS-in-JS by using glamor-glamorous.


examples/glamorous-tailwind/src/index.js
```diff
@@ -0,0 +1,31 @@
...
+// Render your app
+if (typeof document !== 'undefined') {
+  const render = Comp => {
+    ReactDOM.hydrate(
+      <AppContainer>
+        <Comp />
+      </AppContainer>,
+      document.getElementById('root'),
+    )
+  }
```
EmilTholin: Awesome stuff!
Could you change this to [reflect the other examples](https://github.com/nozzle/react-static/blob/master/examples/basic/src/index.js#L13-L21)?

EmilTholin:
Really cool stuff.
I have no Tailwind experience, but do you think the [errors in the build](https://travis-ci.org/nozzle/react-static/builds/311809170?utm_source=github_status&utm_medium=notification) can be fixed easily?

examples/glamorous-tailwind/src/index.js
deden: ok, done. Sorry, I missed that update.

deden:
tailwind.js config file was created by Tailwind-cli utility. Can we just disable it with `/* eslint-disable */` ?

EmilTholin:
Yeah, I don't think that will be a problem. You could alternatively run `yarn test --fix` and see if `eslint` can change it automatically.

viperfx:
@deden I have been trying to follow this example template but use it with styled components. Could you give a bit of info about how glamour is injecting tailwind and how the css is actually served up? I could not figure out by reading the code.

PR #226, opened 2017-12-08
Added a website built with React Static.

hql287:
This is the website that I've built for my latest project: https://github.com/hql287/Manta

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,7 @@ React-Static is a minimalistic framework for server-rendered React applications 
 ## Sites Built with React-Static
 - [Nozzle.io](https://nozzle.io)
 - [Timber.io](https://timber.io)
+- [Manta.life](https://manta.life)
 
 ## Examples and Templates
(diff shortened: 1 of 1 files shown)


EmilTholin:
Your site is incredibly beautiful!

Is there any reason why you are not extracting your css? There is a slight flickering when loading the page, presumably because you inject your styles when the JavaScript bundle is loaded.


flickering_manta_life

hql287:
@EmilTholin : Thank you for the kind words. Yes, I noticed the issue, just was a little busy and didn't have the time to fix it. I'm thinking of putting up a spinner while waiting for the JS to load. What do you think?

EmilTholin:
I haven't seen you Webpack config so I can't say for sure, but I think it will be resolved if you [extract your css like in the default css rule](https://github.com/nozzle/react-static/blob/master/src/webpack/rules/cssLoader.js#L44-L51).

PR #229, opened 2017-12-09
Add the plugin to the sass example

### LINKED ISSUE #228: The Sass test example is missing the plugin
Get an error for when testing the sass example.
image

I have the solution and will push a quick fix PR

talves:
### Is this a bug report? 

Yes
 
Fixes and closes #228

### DIFF
--- a/examples/sass/static.config.js
+++ b/examples/sass/static.config.js
@@ -1,42 +1,40 @@
-import axios from "axios";
-import ExtractTextPlugin from "extract-text-webpack-plugin";
+import axios from 'axios'
+import ExtractTextPlugin from 'extract-text-webpack-plugin'
 
 export default {
   getSiteProps: () => ({
-    title: "React Static"
+    title: 'React Static',
   }),
   getRoutes: async () => {
-    const { data: posts } = await axios.get(
-      "https://jsonplaceholder.typicode.com/posts"
-    );
+    const { data: posts } = await axios.get('https://jsonplaceholder.typicode.com/posts')
     return [
       {
-        path: "/",
-        component: "src/containers/Home"
+        path: '/',
+        component: 'src/containers/Home',
       },
       {
-        path: "/about",
-        component: "src/containers/About"
+        path: '/about',
+        component: 'src/containers/About',
       },
       {
-        path: "/blog",
-        component: "src/containers/Blog",
+        path: '/blog',
+        component: 'src/containers/Blog',
         getProps: () => ({
-          posts
+          posts,
         }),
         children: posts.map(post => ({
           path: `/post/${post.id}`,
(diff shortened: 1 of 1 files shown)


tannerlinsley:
Hold on. Why is this needed if we already have https://github.com/nozzle/react-static/blob/master/src/webpack/webpack.config.prod.js#L34?

talves:
That is a very good question. Let me see if I can see the reason it is not being recognized.

talves:
@tannerlinsley It also works if I change the code to:

`config.plugins = config.plugins`

something weird here on windows maybe?

tannerlinsley:
Yeah I don't see why it wouldn't work without it.

talves:
It works fine for the build, but does not on start, so it seems like a devServer issue. I will have to come back to it when I have some time.

tannerlinsley:
Well that would mean it's something to do with the style loader then,
right? No styles are extracted in dev anyway, right?

talves:
Ok, the reason it is loading for `prod` is because you were right above the plugin is being added in the `webpack.config.prod.js`
For `dev` it is **NOT** as you can see here https://github.com/nozzle/react-static/blob/master/src/webpack/webpack.config.dev.js#L34-L47
The plugin is then missing when we run the devServer (`stage === 'dev'`) with this configuration for the sass example.

I fixed this PR the way I would like to see it loading.

The question is: Are you going to maintain a sass rule in the core webpack config?

talves:
I am not sure if this has been tested, but fixing the error does not fix the production build.
Am I the only one who sees the webpack being called twice for the configs? If you run a build on the basic example, two sets of styles get created, although this is not causing the issue here.

chrisfo:
Could even be done a little like this (a little cleaner?):
```javascript
          {
            test: /\.s(a|c)ss$/,
            use: (stage === 'prod') ?
                ExtractTextPlugin.extract({
                  fallback: 'style-loader',
                  use: [
                    {                    
                      loader: 'css-loader',
                      options: {
                        minimize: true,
                        sourceMap: false
                      }
                    },
                    {
                      loader: 'sass-loader'
                    }                  
                  ]
                })
                :
                [
                  {loader: 'style-loader'},
                  {loader: 'css-loader'},
                  {loader: 'sass-loader'}
                ] 
          },

```

tannerlinsley:
I just pushed some changes.  Let me know what you think.

talves:
Yep, that would work.

tannerlinsley:
Important to note here that you need to use the `stage` variable.

talves:
I was thinking about that when I did the comment above, but thought it was directly tied to `process.env.NODE_ENV === 'development'`.

Good to know, Thanks.

PR #230, opened 2017-12-09
Pass the webpack stats to renderToHtml method so we can extract js chunks into html

rileylnapier:
The existing Dynamic Import didn't support SSR.  It meant the rendered html was always blank.  Using react-universal-component, we can SSR these dynamic imported chunks :)

I updated the example for "dynamic import" ~ if we want a separate example instead let me know?

Thanks


.gitignore
```diff
@@ -6,7 +6,7 @@ coverage
 
 # production
 dist
-lib
+
```
rileylnapier: whoops

tannerlinsley:
Looking good. Just double check nothing is getting committed or changed that doesn't need to be.

tannerlinsley:
Also I'm not sure why it's showing changes to the readme. Make sure you pull in any updates from master

rileylnapier:
yeah something got janky with the readme... I tried to recopy it from master but must have messed up.

Question.  Should this be a new example or updating old dynamic import one?  Might be useful to have and “advanced dynamic import” example

EmilTholin:
@rileylnapier [The existing Option 1](https://github.com/nozzle/react-static/blob/master/examples/dynamic-imports/src/containers/About.js#L22-L37) doesn't work with SSR for you?

EmilTholin:
It should probably be moved to the constructor, since `componentDidMount` in not run in SSR. Don't you think that would be an easier solution? Again, I might be overlooking something here.

```javascript
constructor (...args) {
  super(...args)
  if (typeof document === 'undefined') {
    const { MyComponent } = require('../components/MyComponent')
    this.state = { MyComponent };
  } else {
    this.state = { MyComponent: null }
  }
}
// /////////////////////////////////////////////////////////////////////////
// // Option 1
// // Behaviour:
// //   - Loaded component on static build
// //   - UnLoaded component when react rehydrates
// //   - Loaded component when dynamic import resolves
componentDidMount() {
  import('../components/MyComponent')
    .then(({ MyComponent }) => this.setState({ MyComponent }))
    .catch(error => this.setState({ MyComponent: Failed }))
}
```

rileylnapier:
dynamic imports don't render anything with SSR.  you can show this by running the existing "dyanmic imports" example, "npm run build && npm run serve" then go to /about page.  it will always show the "loading" first since the page doesn't have any ssr.

EmilTholin:
Yes, that's because [Option 1](https://github.com/nozzle/react-static/blob/master/examples/dynamic-imports/src/containers/About.js#L22-L37) is commented out in the starting `dynamic imports` example. If you change to my previous comment, you will have `MyComponent` rendered in SSR.

rileylnapier:
the problem in option 1 is:

`UnLoaded component when react rehydrates`

rileylnapier:
if anything... we don't have to include this in the "examples" ~ but i would still require passing of the client/prod webpack stats to the renderToHtml function

EmilTholin:
Gotcha. Yeah, it's a very powerful addition. Then the user can even use e.g. [React Loadable](https://github.com/thejameskyle/react-loadable) and other alternatives as well.

`react-static` is very much Tanners baby, but just giving my two cents: Wouldn't it then be cool with a separate code splitting example, and a separate `dynamic-imports`?

rileylnapier:
Yeah I think I’ll do another example “advanced dynamic import with ssr”

I originally tried to do this with react-loadable but couldn’t get it to work.  Also that repo has issues turned off so I couldn’t see if anyone else in the past had figured out the issue or ask any questions.  Super odd to me... so I went with react-universal-component instead.

tannerlinsley:
I am perfectly fine with the stat additions. I agree that this should
likely be a different example entirely. I would also be interesting in
discussing the ramifications of sprin supporting this in the core. Is it a
90% use case? Would it hurt users who dont utilize it?

rileylnapier:
Sprin supporting in its core?  Typo?

Do you mean adding this to the core library?  I would support that as the config spans a few different places and because of that it becomes fragile.

Either that or chat about a pluin system?

Also, on a separate note @tanner,  can I schedule some time with you to explain how my company is planning on using the library and chat with you on future plans/roadmap for react-static?

EmilTholin:
Yeah, it would be cool with some sort of opt-in code splitting HOC in core.

```javascript
<CodeSplit withRouteProps loading={() => <div> Loading... </div>}> {/* ... */} </CodeSplit>
```

tannerlinsley:
Agreed. Gatsby does code splitting but it's extremely baked in to the setup
and configuration. If we could allow the same technical support for it
without breaking anything, that would be fantastic. It may not be automatic
like Gatsby, but it would cover that small percentage of sites that may
have massive app bundles. Honestly a lot of the code in that example looks
scary to me right now. Not sure if that's because it's foreign to me or if
it feels fragile.

tannerlinsley:
@rileylnapier feel free to jump on the slack org and DM me.

EmilTholin:
Yeah, code splitting with SSR is pretty scary for all alternatives that exist. The only addition to core is exposing the `prodStats` to `renderToHtml`, and the Webpack stats has to be used by other popular alternatives like e.g. [React Loadable](https://github.com/thejameskyle/react-loadable#mapping-loaded-modules-to-bundles). 
So the fact that the example is complex isn't really an issue for `react-static` itself, I would argue. Exposing `prodStats` will not clash with a future potential code splitting solution in core either.

tannerlinsley:
Can either of you see a nice way to achieve code splitting out of the box
for all users using this approach?

EmilTholin:
React Loadable, which is the least obtrusive code splitting library I know of, still forces you to render  your app in a certain way:

```javascript
// src/index.js
import React from 'react';
import ReactDOM from 'react-dom';
import Loadable from 'react-loadable';
import App from './components/App';

window.main = () => {
  Loadable.preloadReady().then(() => {
    ReactDOM.hydrate(<App/>, document.getElementById('app'));
  });
};
```
And I'm not entirely sure it works with hot module reloading like `react-static` does, so I think @rileylnapier's example would be a great intermediary step until we find something obtrusive that can be added to core.

examples/dynamic-imports/static.config.js
```diff
@@ -36,4 +58,61 @@ export default {
...
+          renderMeta.scripts.map(script => <script type="text/javascript" src={`/${script}`} />)}
+      </Body>
+    </Html>
+  ),
+  webpack: (config, { stage, defaultLoaders }) => {
+    if (stage === 'node') {
+      config.externals = externals
+
+      config.plugins.push(
+        new webpack.optimize.LimitChunkCountPlugin({
```
rileylnapier: Do you all see an issue with always doing this?  If not, we could bake a bunch of the Webpack config into core and nobody would really notice

examples/dynamic-imports/static.config.js
```diff
@@ -36,4 +58,61 @@ export default {
...
+          maxChunks: 1,
+        }),
+      )
+    }
+
+    if (stage === 'prod') {
+      config.output.filename = 'app.[chunkHash:6].js'
```
rileylnapier: Also... I noticed you all do app.[hash] vs chunkhash 

Any reason?  I like the chunkhash so if some bundles don’t change I won’t need to reupload

examples/dynamic-imports/static.config.js
```diff
@@ -36,4 +58,61 @@ export default {
...
+      config.output.chunkFilename = '[name].[chunkHash:6].js'
+
+      config.plugins.push(
+        new webpack.optimize.CommonsChunkPlugin({
+          name: 'bootstrap',
```
rileylnapier: This also could be baked into core

examples/dynamic-imports/static.config.js
```diff
@@ -1,5 +1,27 @@
...
+import path from 'path'
+import webpack from 'webpack'
+import { ReportChunks } from 'react-universal-component'
+import flushChunks from 'webpack-flush-chunks'
+
+const resolve = p => path.resolve(__dirname, p)
+const nodeModules = resolve('./node_modules')
+
+// for SSR of dynamic imports
+const externals = fs
```
rileylnapier: I’m not entirely sure why we can’t use webpack- node-externals but we could move this function to core as well

rileylnapier: I found this pattern digging in the issues of react-universal-component

rileylnapier: We could just add a config that says whitelist external node modules?

examples/dynamic-imports/static.config.js
```diff
@@ -36,4 +58,61 @@ export default {
...
+      <Head>
+        <meta charSet="UTF-8" />
+        <meta
+          name="viewport"
+          content="width=device-width, initial-scale=1.0, minimum-scale=1.0, maximum-scale=1.0"
+        />
+      </Head>
+      <Body className="slug-home">
+        {children}
+        {renderMeta.scripts &&
```
rileylnapier: If we move all the other things to core... then to the user you just need to update renderToHtml and Document... which I feel is much less scary

EmilTholin:
I personally think that it is good as it is right now: Expose `stats` to the user in `renderToHtml`, and let them do whatever they want with it, e.g. use `react-universal-component` like you've done in your great example. I don't think we should move all that to core and have `react-static` lock into that.

rileylnapier:
Fair.  I’ll work on fixing the readme and adding new example so we merge

tannerlinsley:
Those changes you're working on should suffice.  Can't wait!

Issue #210, opened 2017-12-01
Link to anchor doesn't work


DenisNeustroev:

EmilTholin:
Hi @DenisNeustroev! More information is needed in order for anyone to help you. Do you have a minimal reproducible example?

denis-sokolov:
# Workaround
For `#foo` and external links use a simple `<a href>` over a `Link`.

# Analysis 
Here’s the simplest example:

```jsx
import { Link } from "react-static";
<div>
    <Link to="#foo">Go to foo broken</Link>
    <a href="#foo">Go to foo working</a>
    <div style={{ height: "150vh" }} />
    <div id="foo">Foo</div>
</div>
```

Even absolute links do not work for me: `<Link to="http://example.com">Example</Link>` changes the url to `http://localhost:3000/http://example.com`. I run react-static 4.4.0, react-router-dom 4.2.2, react-router 4.2.0.

It seems to be [a design decision in react-router](https://github.com/pshrmn/react-router/blob/7309a507bda66c24411a4a722df4015a8e9db0c9/packages/react-router-dom/modules/Link.js#L73). After all, for anchor links and external links, a `Link` [has no functionality](https://reacttraining.com/react-router/web/api/Link) over an `<a>`. So a current workaround is to use `<a>` for anchor and external links.

In my opinion, it’s bad user experience. At least when I import `Link` from `react-router`, I can expect it to be related to the routing, so it’s for smart router navigation. I would fix it in react-router, but at least it’s understandable. But when I import `Link` from `react-static`, the relation to routing is lost entirely, and now the Link is pure broken.

As a fix, I would wrap the component in a simple component of our own:

```jsx
import { Link, LinkProps, NavLink, NavLinkProps } from 'react-router-dom';

function isRoutingUrl(to: LinkProps['to']) {
  if (typeof to !== 'string') return true;
  return !to.match(/^#/) && !to.match(/^[a-z]{1,10}:\/\//);
}

export function Link(props: LinkProps) {
  if (isRoutingUrl(props.to)) return <Link {...props} />
  return <a href={props.to}>{props.children}</a>;
}

export function NavLink(props: NavLinkProps) {
  if (isRoutingUrl(props.to)) return <NavLink {...props} />
  return <a href={props.to}>{props.children}</a>;
}
```

tannerlinsley:
Yep. You'll have to roll your own Link component or just use a standard a href hash.

denis-sokolov:
[...]

PR #231, opened 2017-12-10
Fix non-routing Link component usages (#210)

denis-sokolov:
#210

### DIFF
--- a/src/links.js
+++ b/src/links.js
@@ -1,0 +1,20 @@
+import React from 'react'
+import {
+  Link as ReactRouterLink,
+  NavLink as ReactRouterNavLink,
+} from 'react-router-dom'
+
+function isRoutingUrl (to) {
+  if (typeof to !== 'string') return true
+  return !to.match(/^#/) && !to.match(/^[a-z]{1,10}:\/\//)
+}
+
+export function Link (props) {
+  if (isRoutingUrl(props.to)) return <ReactRouterLink {...props} />
+  return <a href={props.to}>{props.children}</a>
+}
+
+export function NavLink (props) {
+  if (isRoutingUrl(props.to)) return <ReactRouterNavLink {...props} />
+  return <a href={props.to}>{props.children}</a>
+}
--- a/src/index.js
+++ b/src/index.js
@@ -11,8 +11,6 @@ import { pathJoin } from './shared'
 
 // Proxy React Router
 export {
-  Link,
-  NavLink,
   Prompt,
   Redirect,
   Route,
@@ -20,6 +18,8 @@ export {
   matchPath,
   withRouter,
 } from 'react-router-dom'
+
+export { Link, NavLink } from './links'
 
 // Proxy Helmet as Head
 export { Helmet as Head }


src/links.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react'
+import {
+  Link as ReactRouterLink,
+  NavLink as ReactRouterNavLink,
+} from 'react-router-dom'
+
+function isRoutingUrl (to) {
+  if (typeof to !== 'string') return true
+  return !to.match(/^#/) && !to.match(/^[a-z]{1,10}:\/\//)
```
EmilTholin: Will this work when using the [`hash` version of the Router](https://github.com/nozzle/react-static#router)?

denis-sokolov: Good point, I haven’t tried it.

It should work, as the user of the components should still provide a proper, non-fragment URL, it is the responsibility of `history`/`react-router` to encode the output in whatever format they currently need.

denis-sokolov: I can try and verify it, maybe tomorrow.

EmilTholin:
The [documentation for `Link` and `NavLink`](https://github.com/nozzle/react-static#link-and-navlink) should probably be updated as well to indicate that they are custom `react-static` components now, since `react-router` [doesn't support hash links](https://github.com/ReactTraining/react-router/issues/394#issuecomment-252511289).

It might also cause some confusion that a hash `NavLink` will not have an `active` class or `aria-current` attribute added to it when it is active.

src/links.js
denis-sokolov: No meaningful BC breakages, here are the findings:

1. `<Router type="hash">` and regular links keep working. `<Link to="about">` matches our `isRoutingUrl` check and thus gets passed through to react-router, which renders a `<a href="#/about">`.

2. `<Router type="hash">` and links that manually have the hash route kind of work. `<Link to="#/about">` does not match our `isRoutingUrl` check and gets rendered with the newly introduced shortcut into a `<a href="#/about">`. react-router handles this links, so it responds to user interaction and changes routes, but it does not have `aria-current`. We could attempt to distinguish `to="#/about"` as a non-fragment link, but technically ids [can start with a slash](https://www.w3.org/TR/2011/WD-html5-20110525/elements.html#the-id-attribute). The usage `<Link to="#/about">` is already kind of wrong, and with proper history support [nearly universal](https://caniuse.com/#feat=history), I don’t think we have to worry about this.

3. `<Router type="hash">` and fragment links are still broken after the latest changes. `<Link to="#subheading">` gets rendered with the newly introduced shortcut into a `<a href="#subheading">`, but is still handled by react-router to navigate to `#/subheading`, ignoring the fragment link. This [was already broken](https://github.com/ReactTraining/react-router/issues/394) with the react-router `<Link>`, and is still broken with our manually rendered `<a>`. Nothing to see here.

denis-sokolov:
> The documentation for Link and NavLink should probably be updated as well to indicate that they are custom react-static components now, since react-router doesn't support hash links.

If a package user decides that for hash links they need a manual `<a>`, good for them. They control their fragment links, I am happy for them. This change was meant to be invisible, to stop counter-intuitive behavior of a broken `<Link>`. It’s not meant to canonically replace all `<a>` in users’ applications. Thus, I don’t think it warrants a mention in the documentation.

> It might also cause some confusion that a hash NavLink will not have an active class or aria-current attribute added to it when it is active.

Hypothetically, yes. But those are extra features that someone with an understanding of routing should expect, but then they should also expect them to not work. Still, you’re right, so I’ve added a warning in #232.

PR #234, opened 2017-12-11
Add David Dependancy Status n remove Github Stars

gagan0723:


tannerlinsley:
We like to keep the github stars on there for the occasions that the readme shows up on sites other than github (eg. https://www.npmjs.com/package/react-static)

PR #237, opened 2017-12-13
Update README.md

DanFessler:
Adding Manticore Games to "Sites Built with React-Static" section


EmilTholin:
Oooh, that an AMAZING website!

I think you can get away with maybe shaving off ~70% off `blogheadere.59ab32c8.jpg` and `teampicture.jpg`. 5.2 MB is quite heavy for mobile.

PR #241, opened 2017-12-15
Added Firebase Authentication example

deden:
Feel free to ignore this pull request if you want to, although it might be a useful addition to the examples :)


tannerlinsley:
Would you like to add it to the list of Example in the main readme too?

deden:
OK, I've added it to the list of example in the main readme.

PR #248, opened 2017-12-19
Allow hooking into the beginning of the build process

ChrisNLott:
This PR adds an async function that is called at the beginning of the build process. 

We are using this hook to fetch data that is used during the build process. Is this the right place to handle this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,12 +243,15 @@ export default {
     public: 'public' // The public directory (files copied to dist during build)
   },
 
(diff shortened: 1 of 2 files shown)


tannerlinsley:
I'm assuming you tried to use the get routes function first?

ChrisNLott:
@tannerlinsley can we assume that function will always run at the beginning of the process?

tannerlinsley:
Yes. It will always run before a build :)

PR #255, opened 2017-12-21
Prefetch when route for getRouteProps component change

EmilTholin:
### Is this a bug report?
 
Yes. https://github.com/nozzle/react-static/issues/254
 
 ### Environment
 
 1. `react-static -v`: 4.7.1
 2. `node -v`: 8.9.0
 3. `yarn --version`: 1.3.2
 ### Steps to Reproduce
 
 1. Create a new `basic` example with `react-static create`
 2. Use the same component with the `getRouteProps` HOC for more than one route (like e.g. https://github.com/nozzle/react-static/issues/254)
 3. Start the dev server and navigate between these routes that have the same component
 
 ### Expected Behavior
Should get the route props when the route change

 ### Actual Behavior
The loading spinner appears and `Warning: getRouteProps could not find any props for route: /<your_route_with_the_same_component>.` is shown in the console.
There is currently no logic to fetch the route props when the `match` changes for the same component. This results in `loaded` being set to `true`, but `initialProps` not being set for this new route, so the loading spinner is shown indefinitely.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -240,21 +240,31 @@ export function getRouteProps (Comp) {
     state = {
       loaded: false,
     }
-    async componentWillMount () {
+    componentWillMount () {
       if (process.env.REACT_STATIC_ENV === 'development') {
-        const { pathname, search } = this.context.router.route.location
-        const path = pathJoin(`${pathname}${search}`)
-        await prefetch(path)
-        if (this.unmounting) {
-          return
+        this.loadRouteProps()
+      }
+    }
+    componentWillReceiveProps (nextProps) {
+      if (process.env.REACT_STATIC_ENV === 'development') {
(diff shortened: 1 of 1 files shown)


tannerlinsley:
Awesome :)

miltonbecker:
I ran into the same problem and tried to test this PR to no joy.
I did `npm run build` then `npm link`. Then, on my project's folder, I did `npm link react-static`.
I got a console error on Chrome saying `_this5 is not defined`.

How did you test it?

EmilTholin:
My setup was a bit silly.
I used `react-static create` to create a new `basic` example, removed `react-static` from `node_modules` and clone my fork of `react-static` instead and ran `yarn build`.

Issue #256, opened 2017-12-21
4.7 broke backward compatibility by dropping support for Node 6


denis-sokolov:
My app marks `react-static^4.4.0` as a dependency. I think up to 4.6 it worked seamlessly. Today with the release of react-static 4.7 my build tools failed. Upgrading a required engine is not a backwards-compatible change.

Did you consider upgrading a major version number to 5?

#251

tannerlinsley:
Technically React Static uses internal node tooling that is only available at node 7.5.0, so the fact that you were successfully using React Static with a lower node version is strange. What version of node were you using?

tannerlinsley:
To elaborate more, we did this because we were getting many users attempting to use React Static with an incompatible version, and we simply didn't define what that version was (or know what it was honestly).  So if you can confirm that a lower version works successfully, I will gladly update the minimum version :)

denis-sokolov:
On my local machine, I have node@6.10.2, npm@5.6.0, and run the development server and the build with react-static@4.4.0 right now.

My CI, node@6.12.2 with npm@3.10.10, ran a build command with react-static@4.6.0 just a day ago, 2017-12-20 12:34 UTC.

Same CI, node@6.12.2 with npm@3.10.10, failed to run a build command with react-static@4.7.1 because of #251:

```
TypeError: Object.values is not a function
     at _callee$ (.../node_modules/react-static/lib/static.js:178:32)
```

Whilst react-static did not specify which Node version it required, given that Node 6 worked, and it is an actively supported version, the requirement to run Node 7 comes out of nowhere and breaks the dev processes.

For me it’s trivial to pin the react-static at 4.6 for now, but I care for react-static to succeed long-term. One step would be to ensure we don’t break the processes for the wider community. I would advise to immediately release a 4.7.3, which would support Node 6. Then bump a major version number and move to Node 7 if it is necessary.

denis-sokolov:
(That said, I don’t see value in dropping support for an actively supported, LTS version of Node that has been release merely a year ago. The only newer LTS version has been released just two months ago! It is inconceivable to expect slower folk to already have it, and so not supporting Node 6 limits the user base considerably.)

derekperkins:
@denis-sokolov I see your point, though I think LTS isn't as important here since there isn't actually a node server serving pages. It's fairly trivial to have a CI use a later image just for building the site.

[...]

PR #259, opened 2017-12-23
Sanity build in tests

denis-sokolov:
Although we don’t have tests in the project, an easy start is a single sanity run. This PR does that and gets easy wins to ensure breaking releases are not released.

We build the minimal website with react-static and see that it does not fail.

On top of that, we ensure the test is run in all different Node versions so that older Node versions don’t stop breaking out of nowhere, bringing BC breakage, like #256.


src/static.js
```diff
@@ -37,7 +37,7 @@ export const exportRoutes = async ({ config, clientStats, cliArguments }) => {
       }
 
       // Loop through the props
-      Object.values(route.initialProps).forEach(prop => {
+      Object.keys(route.initialProps).map(k => route.initialProps[k]).forEach(prop => {
```
denis-sokolov: Ignore this line in this diff. It is part of the PR #258 that this PR depends on. Once that is merged, this will disappear from the diff.

tannerlinsley:
This is fantastic!

PR #260, opened 2017-12-23
Remove package-lock.json, fixup 58610df

denis-sokolov:
Two lock files, package-lock.json and yarn.lock run a high risk
of different people and CI environments having different dependencies,
defeating the purpose of the lock files.

In addition, a single lock file clearly describes the prefered tool
in the project. Two lock files make it confusing.

The package.json scripts and the project history shows that yarn is
the preferred tool since the beginning, so let us remove the npm
lock file.


tannerlinsley:
Nice catch. Yarn all the way!

PR #262, opened 2017-12-25
Overwrite Router export from react-router

### LINKED ISSUE #205: history prop missing on Router in TypeScript example
I just created a new app using the TypeScript template and have run into a few issues, including the [existing issue](https://github.com/nozzle/react-static/issues/145) related to the TS definition for `react-static-routes`. My other issue is related to the react-router definitions. After I changed to `src/App.tsx`, I noticed the following TS error related to the use of `<Router>`:

```bash
Type '{ children: Element; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<Router> & Readonly<{ children?: ReactNode; }> & Re...'.
  Type '{ children: Element; }' is not assignable to type 'Readonly<RouterProps>'.
    Property 'history' is missing in type '{ children: Element; }'.
```

Since `Router` [expects a `history` prop](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/modules/Router.js#L11) as part of [`RouterProps`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-router/index.d.ts#L78), I'm trying to understand how react-static works without specifying this prop and whether it's possible to use TypeScript more strictly in a react-static app.

corydeppen:
Fixes #205

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -5,15 +5,14 @@
 
 /// <reference types="react" />
 
-declare module "react-static" {
-
-  import * as React from "react";
+declare module 'react-static' {
+  import * as React from 'react';
 
   // Passing on all react-router typings
-  export * from "react-router-dom";
+  export * from 'react-router-dom';
 
   // Passing on helmet typings as "Head"
-  import * as Helmet from "react-helmet";
+  import * as Helmet from 'react-helmet';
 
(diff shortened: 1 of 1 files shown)


tannerlinsley:
Nice!

PR #265, opened 2017-12-25
Turn off linebreak-style ESLint rule

### LINKED ISSUE #264: Prevent ESLint errors when running tests on Windows
Running `yarn test` on Windows displays a stream of `linebreak-style` ESLint errors due to CRLF linebreaks. I've seen numerous GitHub [conversations](https://github.com/airbnb/javascript/pull/1224) about this across various projects, including the note on the [docs page for the ESLint rule](http://eslint.org/docs/rules/linebreak-style#using-this-rule-with-version-control-systems). It seems like a common way to enforce LF linebreaks and prevent linting errors for devs working on Windows is to add a `.gitattributes` file with `* text=auto` and disable the `linebreak-style` rule.

> When text is set to "auto", the path is marked for automatic end-of-line conversion. If Git decides that the content is text, its line endings are converted to LF on checkin. When the file has been committed with CRLF, no conversion is done. ([source](https://git-scm.com/docs/gitattributes#gitattributes-Settostringvalueauto))

Would you accept a PR that will allow the tests to run as expected on Windows?

corydeppen:
Fixes #264

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,3 +1,6 @@
 module.exports = {
   extends: 'react-tools',
-}
+  rules: {
+    'linebreak-style': 0
+  }
+};
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.js text=auto


tannerlinsley:
Thinking more about it now, could you PR the eslint change to the react-tools/eslint-config-react-tools repo instead?

corydeppen:
Sure. I looked at that repo but wasn't sure if you wanted to affect all that used that config or not.

tannerlinsley:
Do you think there will be breaking changes to the other repos?  Doesn’t
seem like it… but I’m not sure.

corydeppen:
It shouldn't break anything, but it could allow for inconsistent line endings to be introduced in repos that use the config. For those projects that use this config, they should add the same `text` attribute set to `auto` in a `.gitattributes` file to keep things the same as they've been.

tannerlinsley:
I'm thinking we either wait on this, or try and figure out a different way that won't require users to add or edit their .gitattributes file. Thoughts?

corydeppen:
For those that don't want to add a `.gitattributes` file, they would need to explicitly specify how they want to treat line endings in their config. This would put the responsibility on each project that uses this config to decide how they view devs on Windows. I don't like the idea of having a strict default that prevents devs on Windows from running, testing, and contributing to open projects.

tannerlinsley:
Not sure what to do here still...

tannerlinsley:
Reopen if and when needed

PR #269, opened 2017-12-26
Add entry for tsx component

### LINKED ISSUE #261: Module not found when using index.tsx in root
I've run into another issue using the TypeScript example with the latest react-static. If I change `index.js` to `index.tsx` I get the following error:

```bash
Failed to rebuild.
multi ./node_modules/react-hot-loader/patch.js ./node_modules/react-dev-utils/webpackHotDevClient.js (webpack)/hot/only-dev-server.js ./src/index.js
Module not found: Can't resolve 'path\to\src\index.js' in 'path\to\node_modules\react-static\lib\node_modules'
```

I feel like this is another example of an issue that may have been flushed out earlier if the TS example wasn't a hybrid of `js` and `tsx` files. I understand it's possible to mix them, but I'm not sure it's the most common scenario, especially for a new TypeScript project.

corydeppen:
Fixes #261

### DIFF
--- a/examples/typescript/index.d.ts
+++ b/examples/typescript/index.d.ts
@@ -1,0 +1,1 @@
+declare module '*.png'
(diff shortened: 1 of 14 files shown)


tannerlinsley:
Thanks!

PR #279, opened 2018-01-05
Add documentation for hot code reloading and LAN access

hellojere:
I have not tested these features in full, but React-Static seems to support hot code reloading for React Components and CSS as well as LAN access from other devices out of the box. Would be good to emphasise these features in the README.md, as they are crucial for many developers.


tannerlinsley:
Thanks!

PR #292, opened 2018-01-11
fix bundle analyzer and allow to turn off hot modules

rileylnapier:
@tanner i sent you slack about this


tannerlinsley:
Nice work :)

PR #312, opened 2018-01-24
Use bluebird Promise.map to limit concurrency

lottamus:
This PR demonstrates a possible fix for https://github.com/nozzle/react-static/issues/311.


tannerlinsley:
I love this idea. Any way to do it without bluebird?

tannerlinsley:
I would vote for this module: https://github.com/zeit/async-throttle

tannerlinsley:
Another thing, I would love for the concurrency to be set at a sensible default, but more importantly be customizable via the config.

lottamus:
It looks like that repo was archived in favor of https://github.com/zeit/async-sema. I also found https://github.com/timdp/es6-promise-pool.

I will add an option to the config to set the concurrency limit. Do you have a suggestion on the default value?

tannerlinsley:
Try 8 for now

tannerlinsley:
FYI, I'm actually building a pooling and throttling mechanism right now. We can probably just use that :)

tannerlinsley:
Oops

lottamus:
Nice! Shoot it over to me when you're done. 

Do you think we should enforce these limits on all the `Promise.all` calls? I was only running out of memory with https://github.com/nozzle/react-static/blob/master/src/static.js#L106, but I could see others that handle writing files could pose an issue too: https://github.com/nozzle/react-static/blob/master/src/static.js#L61 and https://github.com/nozzle/react-static/blob/master/src/static.js#L89

tannerlinsley:
Almost done :) And yes, possibly, but I say we hold off until it becomes a problem first and avoid the pre-op :)

tannerlinsley:
https://github.com/tannerlinsley/swimmer :)

tannerlinsley:
Any progress here? I've added the swimmer library to master and I'm using it in the client to throttle requests

lottamus:
Closing in favor of #325

PR #314, opened 2018-01-25
[WIP] Babel 7, tree shaking

EmilTholin:
_The current versions of the `@babel` packages and `@std/esm` gives rise to a `TypeError: require(...).default is not a function`. It is probably a good idea to wait for `@babel` to come out of beta, since the combination of `@babel` packages and `@std/esm` breaks between minor changes._ 

This PR moves to Babel 7, moves away from `commonjs` to ES6 with `esm` to allow for tree shaking, and replaces the deprecated `latest` preset with the `env` preset.


jeffvandyke:
What is with the no newline at the end of the files? Best practice is to have all lines of a text file end with a new line character, for many reasons, see <https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline>

EmilTholin:
It's a WIP, i'll fix that later, thanks.

JustFly1984:
Please add support of 'extract-hoc-compose/babel', 'extract-hoc/babel' plugins for correct HMR

jeffvandyke:
Just wanted to say thanks for working on this change. That babel env preset should make future maintenance easier, and this is even now a great, well documented, straight-forward project that had everything I wanted for a small website, and I loved being able to adjust the webpack config to support custom loaders for certain file types.

tannerlinsley:
🤘🤘🙏

tannerlinsley:
🎤 Check. 1, 2, 1, 2. What's the status here?

ryami333:
@tannerlinsley Babel 7 isn't out of Beta yet

tannerlinsley:
:(

PR #321, opened 2018-01-28
Fix link to staticconfig.js documentation

tanayseven:
### Is this a bug report?
Yes

 ### Environment
 
The environment details are irrelevant and therefore I'm leaving them blank
 
 1. `react-static -v`: 
 2. `node -v`:
 3. `yarn --version or npm -v`:
 
 Then, specify:
 
 1. Operating system:
 2. Browser and version (if relevant):
 
 
 ### Steps to Reproduce
 
 1. Open the documentation `https://www.npmjs.com/package/react-static`
 2. Click on `Configuration (static.config.js)` under documentation
 3. Nothing happens
 
 
 ### Expected Behavior
 
It should move to the `Configuration (static.config.js)` section on the page
 
 ### Actual Behavior

Nothing Happens when clicked on that link (the Url changes to something but the page does not scroll to the right element)

When clicked on the following `Configuration (static.config.js)`
screen shot 2018-01-28 at 11 23 12 pm

Should move to the following `Configuration (static.config.js)`
screen shot 2018-01-28 at 11 23 28 pm
 
 
 ### Reproducible Demo
 
The above screenshots should be clear enough

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ If you read these docs on `npmjs.com`, they correspond to the [published version
   - [`react-static start`](#react-static-start)
   - [`react-static build`](#react-static-build)
 - [Project Setup](#project-setup)
-- [Configuration (`static.config.js`)](#configuration-staticconfigjs)
+- [Configuration (`static.config.js`)](#configuration--staticconfigjs-)
 - [Components & Tools](#components--tools)
   - [`<Router>`](#router)
   - [`<Routes>`](#routes)


tanayseven:
Why was this commit reverted? What is the right way to fix this issue?

tanayseven:
I just noticed that the change was working on `npmjs.com` but not working on `github.com`. I don't know how that can be fixed.

tannerlinsley:
Neither do I it's because they use different markdown parsers

tanayseven:
I think one way could be by removing the brackets from 
```
Configuration (`static.config.js`)
```
and change it to 
```
Configuration `static.config.js`
```

Issue #311, opened 2018-01-24
exportRoutes should be throttled (and customizable) for larger sites


lottamus:
The issue here is that we are doing a Promise.all to generate HTML for all routes at the same time: https://github.com/nozzle/react-static/blob/18496dc83ca823a1ca0e6b01193aa8468a29c2df/src/static.js#L106-L269

With a large number of routes (1000+), we quickly run out of memory during this process.

We can easily fix this issue by using bluebird's Promise.map http://bluebirdjs.com/docs/api/promise.map.html, which allows you to set a concurrency limit.

The resulting code would look something like:
```javascript
  return Promise.map(
    config.routes,
    async route => {...},
    { concurrency: 5 } // This number could be calculated based on the number of routes?
  );
```

To reproduce, change the [test/minimal-website/static.config.js](https://github.com/nozzle/react-static/blob/master/test/minimal-website/static.config.js#L1) to the below code

```javascript
module.exports.default = {
  getRoutes: function() {
    var routes = [];

    for (var index = 0; index < 10000; index++) {
      routes.push({
        path: '/' + index,
        component: 'src/Home',
      });
    }

    return routes.map(function(route) {
      return Object.assign(route, {
        getProps: function() {
          return routes;
        },
      });
    });
  },
};
```

Then run `yarn test-sample-build`

Result:

```
➜  react-static git:(master) ✗ yarn test-sample-build
yarn run v1.3.2
$ yarn build && (cd test/minimal-website && ../../bin/react-static build)
$ rimraf lib && babel src --out-dir lib
src/RootComponents.js -> lib/RootComponents.js
src/commands/build.js -> lib/commands/build.js
src/commands/create.js -> lib/commands/create.js
src/commands/index.js -> lib/commands/index.js
src/commands/start.js -> lib/commands/start.js
src/index.js -> lib/index.js
src/links.js -> lib/links.js
src/shared.js -> lib/shared.js
src/static.js -> lib/static.js
src/utils.js -> lib/utils.js
src/webpack/index.js -> lib/webpack/index.js
src/webpack/rules/cssLoader.js -> lib/webpack/rules/cssLoader.js
src/webpack/rules/fileLoader.js -> lib/webpack/rules/fileLoader.js
src/webpack/rules/fontsLoader.js -> lib/webpack/rules/fontsLoader.js
src/webpack/rules/index.js -> lib/webpack/rules/index.js
src/webpack/rules/jsLoader.js -> lib/webpack/rules/jsLoader.js
[...]

PR #325, opened 2018-01-29
Use poolAll whenever outputting multiple files

lottamus:
This PR demonstrates a possible fix for #311.


tannerlinsley:
Wow. Looks fantastic! I think we should tie limits into the config. Thoughts? Check out the utils file where the config is built. As for the stringify, it only happens once. But what could we do to mitigate that? 

Also, how do you like swimmer?!

lottamus:
Add `outputFileRate` to the config object with a default of 10 and added both `prefetchRate` and `outputFileRate` to the README.

It was super easy to add swimmer's `poolAll` function for this use-case.

tannerlinsley:
Awesome!

PR #332, opened 2018-02-01
API Reference: Remove duplicate devServer

hellojere:
The information for devServer config object is repeated twice in the API Reference docs. 

I also would personally would like to see the default port 3000 on the example - helps greatly when searching how to change this setting. Alternative option is to mention something in the line of `Change from default port 3000` in a comment line.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -309,21 +309,6 @@ export default {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


tannerlinsley:
Awesome!

PR #335, opened 2018-02-01
styled-components config siteData functional call fix

hellojere:
Wrong function called in static.config.js for styled-components demo to get site data, changed from `withSiteData` to `getSiteData`


tannerlinsley:
Thanks!

PR #361, opened 2018-02-08
Add missing typing for withSiteData

sasuomi:
Fixes
```
[ts] Module ''react-static'' has no exported member 'withSiteData'.
```
when trying to
```js
import { withSiteData } from 'react-static'
```

Although other type definitions could use a little love as well, IMO this is pretty important to add considering the error shows up immediately after initializing a new project with the [official TypeScript template](https://github.com/nozzle/react-static/tree/master/examples/typescript).

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -18,6 +18,7 @@ declare module 'react-static' {
 
   export function getRouteData(comp: any): any
   export function getSiteData(comp: any): any
+  export function withRouteData(comp: any): any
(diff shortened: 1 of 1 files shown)


tannerlinsley:
You could just remove the two get- methods and add the withRouter start while you're at it :)

sasuomi:
Sure! Done

tannerlinsley:
Awesome!

PR #365, opened 2018-02-09
Add Cryptagon to sites made with React-Static

bernatfortet:
I just added a new entry to the list of sites made with react-static


tannerlinsley:
Awesome!

PR #373, opened 2018-02-13
Basic Algolia example

hellojere:
Added a basic example of how to include Algolia search to react-static. Based on the `basic` example and uses the same data with a working search route added to the navigation. 

The required changes for this to appear when starting a new project with `react-static create` are not included.


tannerlinsley:
The example will automatically appear as a `react-static create` option since it's in the examples directory :) Nice Work!

PR #381, opened 2018-02-15
Add Typetalk to sites made with React-Static

johnykei:
I added Typetalk site to the list of sites made with react-static.
React Static is really cool! Thank you!!!


tannerlinsley:
Thanks! My pleasure.

PR #385, opened 2018-02-15
react-static create CLI arg handler

eruby94:
The new (optional) react-static create command will allow for template and name to be passed as args.

Example:
```bash
react-static create --name=some-project-name --template=basic
```

When a particular arg is omitted, the user will be prompted as currently done for that input. If both are provided, and the template is valid, then both prompts will be skipped.

Further specifics may be observed in the code.


src/commands/create.js
```diff
@@ -11,29 +11,64 @@ import { ChalkColor } from '../utils'
...
+    if (args[i].indexOf('--name') === 0) {
+      nameInput = args[i].substring(7)
+    } else if (args[i].indexOf('--template') === 0) {
+      const templateInput = args[i].substring(11)
+      if (examples.indexOf(templateInput) !== -1) {
+        template = templateInput
+      }
+    }
+  }
+  if (typeof nameInput === 'undefined') {
```
tannerlinsley: `if (nameInput) {` should suffice here

src/commands/create.js
```diff
@@ -11,29 +11,64 @@ import { ChalkColor } from '../utils'
...
+    prompts.push({
       type: 'input',
       name: 'name',
       message: 'What should we name this project?',
       default: 'my-static-site',
-    },
-    {
+    })
+  }
+  if (typeof template === 'undefined') {
```
tannerlinsley: Same here

src/commands/create.js
```diff
@@ -11,29 +11,64 @@ import { ChalkColor } from '../utils'
...
+  const files = await fs.readdir(path.resolve(__dirname, '../../examples/'))
+
+  console.log('')
+
+  let exampleList = files.filter(d => !d.startsWith('.'))
+  exampleList = ['basic', ...exampleList.filter(d => d !== 'basic'), 'custom']
+
+  const { prompts, nameInput, template } = handleCliArguments(cliArguments, exampleList)
+
+  const shouldPrompt = typeof nameInput === 'undefined' || typeof template === 'undefined'
```
tannerlinsley: `const shouldPrompt = !nameInput || !template`

src/commands/create.js
```diff
@@ -11,29 +11,64 @@ import { ChalkColor } from '../utils'
...
+
+  const answers = shouldPrompt ? await inquirer.prompt(prompts) : {}
+
+  if (typeof nameInput !== 'undefined') {
```
tannerlinsley: `if (nameInput) {`

src/commands/create.js
```diff
@@ -11,29 +11,64 @@ import { ChalkColor } from '../utils'
...
+    answers.name = nameInput
+  }
+  if (typeof template !== 'undefined') {
```
tannerlinsley: `if (template) {`

tannerlinsley:
Now all it needs is documentation 😉

PR #393, opened 2018-02-19
WIP add custom error page

kildareflare:
I thought it would be nice to have a custom error page, for production in particular and that it could work in a similar manner to the 404 page.
I've made a start on this but had to move onto something else. I'll pick it up again shortly.
One thing I was not sure about was how to actually switch in the custom page. Initially I thought this could be done on line 682 of index.js - but could not find an obvious way of doing this. 

Any ideas?

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -678,6 +678,9 @@ class Router extends Component {
 
     if (this.state.error) {
       // Fallback UI if an error occurs
+
+      // TODO If there is a custom error route defined, use that instead
+
       return (
         <div
           style={{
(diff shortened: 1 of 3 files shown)


tannerlinsley:
This leans me towards getting rid of the is404 flag and just having users define a 404 route. Then it's ready enough to allow them to define a 500 route as well.

tannerlinsley:
Concerning the catching of the error, you would simply push the location to 500 using react routers history API.

tannerlinsley:
Moving this PR to "stalled" status.  This is still a cool idea we can keep discussing.

kildareflare:
Hey, this is still on my todo list, got swamped lately. I'll get back to it soon hopefully.

PR #404, opened 2018-02-21
rename gitignore -> .gitignore in various examples

kylemac:
### Is this a bug report?
 
Yes

 ### Environment
 
 
 
 1. `react-static -v`: @current
 2. `node -v`: v8.2.1
 3. `npm -v`: 5.6.0
 
 Then, specify:
 
 1. Operating system: Mac osx
 2. Browser and version (if relevant): N/A
 
 
 ### Steps to Reproduce
 
 
 
 (Write your steps here:)
 
 1. `$ react-static create`
 2. `fill out interactive prompts`
 3. `cd <new-project>`
 
 
 ### Expected Behavior
 
 
 
I expected the `gitignore` files to be functional and properly named. 
 
 
 ### Actual Behavior
 
 
 
a few `.gitignore` files in the various `/examples` directory were without the preceeding `.` in their filename. i couldn't think of an actual reason for this so had to assume some of the examples were based off the same core files that had the mis-named `.gitignore` filename


tannerlinsley:
This is intentional. When attempting to handle/copy/generate them into examples using `react-static create`, they would be ignored themselves or not copied. This is a common solution that is also used in CRA.

tannerlinsley:
To be clear, when using `react-static create` the `.gitignore` file should be there and function as expected.

kylemac:
ah, I didn't realize. I had to abort the `yarn install` bc it seemed to hang. I didn't realize files would be renamed later in the process than that. 

that said, in some `/examples` the file is already named `.gitignore` and when I just ran a test with   `/examples/firebase-auth`, i got this error: 

```
Error: ENOENT: no such file or directory, stat '~/Sites/null/testproject/gitignore'
```

should I PR updates to make them all `gitignore`?

tannerlinsley:
Yeah. Most of them should be already

PR #407, opened 2018-02-22
 add react-hot-loader 4 dep to prismic example

kylemac:
when updating to react-hot-loader 4 in this commit https://github.com/nozzle/react-static/commit/135c499637838cc14f30b779b06a812db2aefcee the `examples/basic-primsmic/package.json` wasn't updated causing that example not to work right away when generating with `react-script create` cli

### DIFF
--- a/examples/basic-prismic/package.json
+++ b/examples/basic-prismic/package.json
@@ -6,6 +6,7 @@
     "prismic-javascript": "^1.5.0-beta.1",
     "react": "^16.0.0",
     "react-dom": "^16.0.0",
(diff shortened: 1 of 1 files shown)


tannerlinsley:
This should work without the local dependency. Does it not? It's installed at the react static level.

kylemac:
you're right. not sure why locally it was barking at me. another swing and a miss PR by me. sorry - just getting familiar with the project and was trying to be helpful

tannerlinsley:
No worries man! I’m just happy to see you actively using it :)

PR #415, opened 2018-02-25
refactor cli interface to use commander lib. add support for remote template URLs

kylemac:
this PR addresses the following: 
- minor cosmetic clean up to `package.json` using https://github.com/HenrikJoreteg/fixpack
- some cosmetic clean up via linting on `src/commands/create.js` 
- update CLI to utilize https://github.com/tj/commander.js/ library to define cmds + options
- some expanded examples within CLI `--help` cmds
- support for remote repo URLs (ssh + https) when using `react-static create --template` opt
- document changes to CLI API in `/docs` 

some notes:
- i wasn't sure what this logic was meant to accomplish, so I omitted it and did not run into any issues. please let me know if it should be adding back: https://github.com/nozzle/react-static/blob/master/src/commands/index.js#L5-L29
- same for this: https://github.com/nozzle/react-static/blob/master/src/commands/index.js#L31-L37

[**edited**: fixed link in second bullet point of "some notes"]


kylemac:
not totally sure why the netlify build fails but I think it has to do with the `gitdocs` dependencies not being properly installed globally on the build machine. `gitdocs build` works properly on my local machine but I can't debug the netlify build server w/o being given ascess

tannerlinsley:
The code in question is necesarry at any entry point of the cli to ensure that Babel-register does not attempt to transform non JavaScript files.

tannerlinsley:
Also you two references to the same code. In your comment. Was that intended?

kylemac:
@tannerlinsley okay, i'll add it back - good to know. also, i meant for the second reference to be this link: https://github.com/nozzle/react-static/blob/master/src/commands/index.js#L31-L37

tannerlinsley:
To the second piece of code, it's to solve nodes cumbersome tendency to not include stack traces with unhandled exceptions, especially given that users can run their own arbitrary code in the node context within the config.

tannerlinsley:
So how do the sub commands get run? If they are entry points themselves , they will need the ignores extensions and rejection catch as well no?

kylemac:
commander delegates the sub-commands based on a filename prefix for you

PR #421, opened 2018-03-01
Use an apostrophe rather than a prime

stephenmathieson:
This patch replaces a prime with an apostrophe (see https://webdesignledger.com/common-typography-mistakes-apostrophes-versus-quotation-marks/). Normally I'd consider this as "not worth the time", but because a prime is used, GitHub's syntax highlighting gets confused and the code is illegible.

----

screen shot 2018-03-01 at 12 57 56 pm

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -694,7 +694,7 @@ class Router extends Component {
             background: 'rgba(0,0,0,0.05)',
           }}
         >
-          <h2>Oh-no! Something's gone wrong!</h2>
+          <h2>Oh-no! Something’s gone wrong!</h2>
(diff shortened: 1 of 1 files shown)


tannerlinsley:
What if we just do something like `{"'"}`?

stephenmathieson:
That might work, but would does intentionally using hacks rather than proper punctuation make sense? IMO it'd be hard to understand _why_ `{"'"}` was there, whereas most folks won't notice the apostrophe (and those who do will appreciate it).

tannerlinsley:
It's not mission critical, so I'm fine with the fix for now.

PR #424, opened 2018-03-02
Update react-static to use webpack4

GeertWille:
Hey,

I did some initial setup to integrate [Webpack 4](https://medium.com/webpack/webpack-4-released-today-6cdb994702d4) into react-static. This is just a WIP PR so it is possible that there are still issues present but we can discuss them here and solve them. Webpack 4 is way faster than Webpack 3. In a specific project we went from 3 minutes to 30s. So check it out ;)


tannerlinsley:
Any update on this?

GeertWille:
I will continue to work on this when I find some time ;)

PR #427, opened 2018-03-03
Add CssHash for extract-css-chunks-webpack-plugin support

Daniellt1:
By the way, would it be sensible to replace `extract-text-webpack-plugin` with `extract-css-chunks-webpack-plugin` in default css loader? In my opinion, `extract-css-chunks-webpack-plugin` seems much better option which could be available in `react-static` out of the box.

### DIFF
--- a/src/static.js
+++ b/src/static.js
@@ -186,6 +186,7 @@ export const exportRoutes = async ({ config, clientStats, cliArguments }) => {
       let head = {}
       let clientScripts = []
       let clientStyleSheets = []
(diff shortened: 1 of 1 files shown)


Daniellt1:
Because decision have been made to try to replace `extract-text-webpack-plugin` with `extract-css-chunks-webpack-plugin` - this PR can be close. `CssHash` going to be added as a part of another PR which going to contain mentioned plugin replacement.

PR #441, opened 2018-03-06
(docs) - fix docs.  (feat) - add key === pathname.

rileylnapier:
documentation for adding custom Route render was wrong.

also, i added the key === pathname for the default component.  

different routes of the same path component need to be re-rendered.  without the key, the router doesn't know to re-render.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -577,11 +577,15 @@ import { Router } from 'react-static'
 import Routes from 'react-static-routes'
 
 // This is the default renderer for `<Routes>`
-const RenderRoutes = ({ getTemplateForPath }) => (
+const RenderRoutes = ({ getComponentForPath }) => (
   // The default renderer uses a catch all route to recieve the pathname
   <Route path='*' render={props => {
     // The pathname is used to retrieve the component for that path
-    let Comp = getComponentForPath(props.location.pathname)
+    let Comp = getComponentForPath(
+      props.location.pathname === '/'
+        ? '/'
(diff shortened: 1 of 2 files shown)


docs/api.md
```diff
@@ -577,11 +577,15 @@ import { Router } from 'react-static'
...
 
 // This is the default renderer for `<Routes>`
-const RenderRoutes = ({ getTemplateForPath }) => (
+const RenderRoutes = ({ getComponentForPath }) => (
   // The default renderer uses a catch all route to recieve the pathname
   <Route path='*' render={props => {
     // The pathname is used to retrieve the component for that path
-    let Comp = getComponentForPath(props.location.pathname)
+    let Comp = getComponentForPath(
+      props.location.pathname === '/'
```
rileylnapier: i think in your codebase you have something called "cleanPath"??

src/static/generateRoutes.js
```diff
@@ -97,7 +97,7 @@ export default class Routes extends Component {
           is404 = true
           Comp = getComponentForPath('404')
         }
-        return Comp ? <Comp {...props} {...(is404 ? {is404: true} : {})} /> : null
+        return Comp ? <Comp key={props.location.pathname} {...props} {...(is404 ? {is404: true} : {})} /> : null
```
rileylnapier: adding this prevents me from having to add a custom render to Routes.  would love to use vanilla `<Routes />`

tannerlinsley:
The implementation of this has changed.

PR #446, opened 2018-03-08
Add options to use extractCssChunks and inlineCss

Daniellt1:
Two config options were added:

1. `extractCssChunks` - replaces default `ExtractTextPlugin` with `ExtractCssChunks`. It enables automatic CSS splitting into separate files by routes as well as dynamic components (usign `react-universal-component`). More information about the [plugin](https://github.com/faceyspacey/extract-css-chunks-webpack-plugin) and [why it is useful as a part of CSS delivery optimisation](https://github.com/faceyspacey/extract-css-chunks-webpack-plugin#what-about-glamorous-styled-components-styled-jsx-aphrodite-etc). Defaults to `false`.
2. `inlineCss`. By using extractCssChunks option and putting code splitting at appropriate places, your page related CSS file can be minimal. This option allows you to inline your page related CSS in order to speed up your application by reducing the number of requests required for a first paint. Default to false.

Worth to mention in the changelog: if project has custom CSS loader with `ExtractTextPlugin`, `extractCssChunks` will cause an error during the build. That could be solved by replacing the plugin manually in custom loader.

### DIFF
--- a/src/static/index.js
+++ b/src/static/index.js
@@ -224,6 +224,8 @@ export const exportRoutes = async ({ config, clientStats }) => {
       let head = {}
       let clientScripts = []
       let clientStyleSheets = []
+      let clientCss = {}
+      let ClientCssHash
 
       let FinalComp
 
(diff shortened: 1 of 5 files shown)


tannerlinsley:
Awesome!

src/static/index.js
```diff
@@ -331,6 +336,15 @@ export const exportRoutes = async ({ config, clientStats }) => {
             {head.link}
             {head.noscript}
             {head.script}
+            {config.inlineCss &&
+              <style
+                key="clientCss"
```
denis-sokolov: This `key` here is not necessary, btw.

denis-sokolov:
When possible, I would like to suggest we minimize the options and keep the main flow the best for all our users. Our description says “It's been carefully designed to meet the highest standards of [..] developer experience”, but this comes close to “Install and then spend 3 hours tuning the options”, which is far from the highest standards.

Besides, every option adds a combinatoric risk of having code flows that we don’t at all use, and we might break then and not notice.

In this case, why would we not enable these options by default? @tannerlinsley

PR #453, opened 2018-03-09
Adding new example on-the-fly-routing

Jonarod:
A new example to show how to create routes directly from the client's UI, without them being listed in static routes.


tannerlinsley:
Looks good. Just needs to pass linting.

PR #460, opened 2018-03-11
Prevent scroll from being stuck

### LINKED ISSUE #449: Hash scrolling rubber-bands when content changes
Scenario:

1. Bet a URL `/example#foo`, scrolled down to the end of the page.
2. Click on a link `#bar`.
3. `onhashchange` a piece of custom code removes `<div id="foo">`, adds a `<div id="bar">` in its place.
4. Screen shows the new div, and the page instantly scrolls to the top of `<div id="bar">`.
5. With no delay start scrolling down.

Expected:

6. Page scrolls down.

Actual:

6. Page scrolls down a small amount and is immediately scrolled to the top of `<div id="bar">`. This repeats for a better part of a second, after which the scroll is released to act naturally.

_I’m sorry to bring the hash scrolling again, you must hate it by now._

denis-sokolov:
Whatever the reason behind the scroll getting to the location too early,
if we do not abort, we will continue doing no-op `window.scroll` operations
for the rest of duration. This means that the user in the meantime can not scroll.

Fixes #449.

### DIFF
--- a/src/utils/scrollTo.js
+++ b/src/utils/scrollTo.js
@@ -49,6 +49,13 @@ export default function scrollTo (element, options) {
         resolve()
         return
       }
+
+      // Sanity check to prevent taking over the scroll once we prematurely got to the element
+      if (start === end) {
+        resolve()
+        return
+      }
+
       raf(step)
     }
     raf(step)


tannerlinsley:
Is it possible to overshoot the exact === ?

denis-sokolov:
I considered to put a comparison with the epsilon:

```js
if (Math.abs(start - end) < 3) {
  scrollTo(end);
  resolve();
  return;
}
```

This would disable very small scrolls in favor of instant bumps. This might be what you want, but for now the PR is as minimally intrusive as possible so that we have the option of merging this immediately.

@tannerlinsley

tannerlinsley:
Sounds good. Make it pass tests and I'll merge

PR #461, opened 2018-03-11
Better logging for getRouteInfo(). No prefetch for undefined routes. Fixes #459

### LINKED ISSUE #459: getRouteInfo() - "Could not load routeInfo for path: undefined"
Two thoughts on `getRouteInfo()`:

The first is that it probably shouldn't trigger an error when the query string of a route changes, as RS doesn't handle routes with query strings. For reference, I'm changing the query string with `history.replace({search})`. 

The second thought is that this function should save the original `path` parameter for this error, rather than logging the cleaned path. Purely for easier debugging.

I'm happy to submit a PR.

mikestopcontinues:
First commit allows for better logging when getRouteInfo can't find routeInfo.

Second commit disables prefetch when path is undefined (e.g., when updating only the query string).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "react-static",
-  "version": "5.1.14",
+  "version": "5.5.13",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


tannerlinsley:
Make sure you run eslint and if possible, prettier on the code so it passes tests.

tannerlinsley:
Nice work!

PR #462, opened 2018-03-11
Fix corner case of hash scroll not happening

### LINKED ISSUE #449: Hash scrolling rubber-bands when content changes

denis-sokolov:
In my codebase that I did not manage to reduce for a demonstration,
RAF would delay the scrollTo call so much that the browser would have already instanly scrolled us to the correct location.
As a result the scrollTo would then be a no-op.

Instead, we now attempt to scroll as soon as possible, synchronously,
and fall back on RAF if that did not work.

Fixup efbaa1618. Fixes #449. Related to #460, but does not make that irrelevant.

### DIFF
--- a/src/client/components/RouterScroller.js
+++ b/src/client/components/RouterScroller.js
@@ -41,24 +41,33 @@ const RouterScroller = withRouter(
       if (!autoScrollToHash) {
         return
       }
-      RAF(() => {
-        if (hash) {
-          const resolvedHash = hash.substring(1)
-          if (resolvedHash) {
-            const element = document.getElementById(resolvedHash)
-            if (element !== null) {
-              scrollTo(element, {
-                duration: scrollToHashDuration,
-                offset: scrollToHashOffset,
-              })
-            }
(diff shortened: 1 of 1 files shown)


tannerlinsley:
I'm curious of this will still solve the problem I was experiencing in gitdocs where the element was found, but the repaint hadn't happened yet. I think it should. 👍

PR #464, opened 2018-03-12
Fix server-side rendering of data with UTF-8 special characters

ViliamKopecky:
### Is this a bug report?
 
Yes
 
 
 
 ### Environment
 
 
 
 1. `react-static -v`: 5.5.14
 2. `node -v`: v8.9.4
 3. `yarn --version or npm -v`: yarn 1.2.1
 
Operating system: macOS

Browser and version (if relevant): Chrome
 
 
 ### Steps to Reproduce
 
 
 
 1. Use specific UTF-8 characters in getData
 2. Load the URL page by browser reload (redirect by AJAX works well)
 3. RS shows 404, because there is an error about unparsable JS code.
 
 
 ### Expected Behavior
 
 
 
Page should load well with browser refresh and show correct data.
 
 
 ### Actual Behavior
 
 
 
404

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,6 +67,7 @@
     "html-webpack-plugin": "^2.30.1",
     "inquirer": "^3.3.0",
     "inquirer-autocomplete-prompt": "^0.11.1",
+    "jsesc": "^2.5.1",
     "match-sorter": "^2.0.2",
     "openport": "^0.0.4",
     "postcss-flexbugs-fixes": "^3.2.0",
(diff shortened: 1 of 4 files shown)


tannerlinsley:
Do you know if this affects build performance?

D1no:
@tannerlinsley would be great if we can merge and publish this. I just encountered the same problem with the german ä, ö, ü characters that are quite common around here ¯\_(ツ)_/¯

https://github.com/nozzle/react-static/pull/464/files#diff-ef1db16ec846c0422615744e3fe14f65R7
`'Ich ♥ Bücher': 'foo 𝌆 bar',`
`here____^`

tannerlinsley:
Sounds like performance is negligible. Merging now :)

PR #473, opened 2018-03-14
Fixed --config arguments

sebastianseilund:
682d35a9c930386133997b86d603b9be85495628 broke the ability to use --config to point to a static.config.js file.

Non-boolean options should include a `<value>` segment in the option commander definition.


tannerlinsley:
Excellent find!

PR #480, opened 2018-03-18
TypeScript: Excluded node_modules in tsconfig.json

chibicode:
I'm pretty sure `node_modules` should be excluded - this was preventing my dev environment to load React-specific types correctly.

[According to the official doc](http://www.typescriptlang.org/docs/handbook/tsconfig-json.html):

> The `"exclude"` property defaults to excluding the `node_modules`, `bower_components`, `jspm_packages` and `<outDir>` directories when not specified.

## If `node_modules` was not excluded

My `code` tag isn't using type definitions from `@type/react`:

screen shot 2018-03-17 at 11 37 05 pm

## If `node_modules` was excluded

My `code` tag uses type definitions from `@type/react`:

screen shot 2018-03-17 at 11 40 54 pm

### DIFF
--- a/examples/typescript/tsconfig.json
+++ b/examples/typescript/tsconfig.json
@@ -31,6 +31,7 @@
   },
   "exclude": [
     "dist/**/*",
-    "public/**/*"
+    "public/**/*",
+    "node_modules"
   ]
 }


D1no:
That is odd. TypeScript usually excludes a set of folders — incl. ‚node_modules‘ — by default.

09983eab-c22a-409b-822c-3dfeb59139ad

However, merging this does not hurt. Before that, can you update the taxonomic in the less android example as well?

chibicode:
@D1no the same documentation page has an example of using `exclude` property, which has `node_modules` and `**/*.spec.ts`:

screen_shot_2018-03-18_at_1_06_09_am

So that made me ask: does `exclude` option **override** the default excludes (`node_modules`, `bower_components`, `jspm_packages`, and `<out-Dir>`), instead of **adding** to the default excludes? I tried to dig in the typescript source code but I couldn't understand figure out if that was the case.

> Before that, can you update the tsconfig in the less antdesign example as well?

Sure, will do that soon.

D1no:
Yeah I think any exclude field knocks out the default. Good catch

tannerlinsley:
Good find!

chibicode:
@tannerlinsley @D1no thanks, just updated the less antdesign example as well. https://github.com/nozzle/react-static/pull/490

PR #482, opened 2018-03-18
Show URL and component path of error during build

ViliamKopecky:
When build fails, I am often missing the URL and component which failed. This enhances the original Error message with info about those things.


tannerlinsley:
Excellent!

PR #483, opened 2018-03-18
Using pretty-error package for nicer error reporting during build

ViliamKopecky:
Before:

screen shot 2018-03-18 at 21 53 36

After:

screen shot 2018-03-18 at 21 52 29

I would like to see this also in watch mode.


tannerlinsley:
Awesome addition!

PR #486, opened 2018-03-19
Cssmodules support with cssModules boolean option in static.config.js

JustFly1984:
I've digged deeper into the source code and found a solution to turn on CSS Modules support. Please consider to update Webpack 4 version to support CSS modules too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ React-Static is a fast, lightweight, and powerful framework for building static-
 - [Typetalk - Chat App for Businesses and Teams](https://www.typetalk.com "Chat App for Businesses and Teams")
 - [This Is Our Home - eCommerce](https://www.tioh.co "This Is Our Home - eCommerce")
 - [Lam Hieu - Personal Website](https://lamhieu.info)
-- [Elsa Salonen - Artist Portfolio](https://elsasalonen.com/) 
+- [Elsa Salonen - Artist Portfolio](https://elsasalonen.com/)
 - [PSD Wizard: On-demand Front-End Coding Service](https://psdwizard.com)
 
 ## Quick Start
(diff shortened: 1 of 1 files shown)


src/static/webpack/rules/cssLoader.js
```diff
@@ -11,10 +11,15 @@ export default function ({ config, stage }) {
...
+          options: config.cssModules
+            ? {
+              importLoaders: 1,
+              modules: true,
+              localIdentName: '[name]__[local]___[hash:base64:5]',
+            } : {
+              sourceMap: true,
+              importLoaders: 1,
+            },
         },
```
tannerlinsley: I would rather not encourage changes to the base loaders, and instead have this done through `config.webpack`. Thoughts?

JustFly1984:
Why not? You are passing config to each base loader anyway, and change
surface is pretty small. I can definitely make it as you want, but for me
it feels right way to utilise static.config.js

Also, I think about webpack 4 implementation, and external changes to
webpack.config could break when you roll it out. Anyway I could miss
something.

JustFly1984:
Tanner, what do you think? Does somebody else has opinion on this matter?

tannerlinsley:
I still feel like the change should live in the config for now. The way I see it, very soon possibly, a "plugin" system is going to come into play, which means it should be utilizing the same API as the user would. I think it should go in a `webpack` declaration in the config.

rachel-church:
I agree that the change should live in the `config.webpack`. It is a simple thing for a user to configure if they need modular css.

For my project I am using a customized css-loader that supports both global files and modular files. In my case setting `cssModules: true` wouldn't be enough.

But I do think that it is a good idea to have an example or at least docs discussing how to add the couple of lines needed to enable modular css support

tannerlinsley:
@rachel-church Hit the nail on the nose. Going to close this, but I fully expect there to be an awesome example showing how to set this up.

PR #493, opened 2018-03-19
Keep user's devServer.watchOptions when running dev server

adambuczek:
Closes https://github.com/nozzle/react-static/issues/479


tannerlinsley:
Nice Work!

PR #498, opened 2018-03-20
Change default devDist to `tmp/dev-server`, fix #497

### LINKED ISSUE #497: building interferes with dev server
With the default configuration `react-static start` and `react-static build` output into the same directory, and this conflict prevents comfortable local development and testing.

https://github.com/nozzle/react-static/blob/86474f41d1501f2a19b253bc1a9c605cb7b94802/src/static/getConfig.js#L24-L30
https://github.com/nozzle/react-static/blob/86474f41d1501f2a19b253bc1a9c605cb7b94802/src/static/getConfig.js#L38

Since b2547a984 we can use the `paths.devDist` option to fix this, but why not have it better by default? Why is the default configuration conflicting one?

denis-sokolov:
This is a quick, untested suggestion. If accepted, we need to review it and verify it works.


tannerlinsley:
Yeah, let's test this with a few projects. Do you think `tmp/dev-server` is a good location vs `devDist`? I'm open to either, as long as it is clean and users can understand what is happening.

denis-sokolov:
This is my personal preference. `tmp` is a clear name for a clear purpose, and we can reuse it for other temporary needs. (Putting the discussion to use OS-level-user/system-wide tmp directories aside)

`devDist` is not clear to me. “Distribution files for development”? Distribution contradicts the development environment in my mind.

tannerlinsley:
Once again, great explanation. Temp dev server is great. Is this ready to
merge you think?

denis-sokolov:
Sorry for no reply. Uh, I wasn’t sure I have covered the parts that matter. In particular, I was unsure about the multiple gitignores. But it seems you’re on top of it.

PR #501, opened 2018-03-20
Disable NODE_ENV=production in build --debug

### LINKED ISSUE #500: `build --debug` silences React warnings
`react-static build --debug` does not help me debug by silencing React warnings. Namely, React 16 has important [SSR HTML mismatch warnings](https://github.com/facebook/react/issues/10085), but `react-static build --debug` does not show them, because `react-static build --debug` still builds with `process.env.NODE_ENV === 'production'`, and React disables all warnings in that `NODE_ENV`.

denis-sokolov:
Fix #500. Also removes React warning about unminified React in production.

I’m not sure you want this as the best solution to #500. After all, this makes it so that `build --debug` is now one step more different from production. At the same time it was super strange that `--debug` does not show, you know, debug information. Finally, maybe we think that non-minification already makes this build far enough from production, so might as well not set NODE_ENV.

This is a judgement call, and I am not sure about this solution, thus this PR is separate from the issue.

### DIFF
--- a/src/commands/build.js
+++ b/src/commands/build.js
@@ -8,7 +8,7 @@ import { copyPublicFolder } from '../utils'
 
 export default async function build ({ config, staging, debug, isCLI, silent = !isCLI } = {}) {
   // ensure ENV variables are set
-  if (typeof process.env.NODE_ENV === 'undefined') {
+  if (typeof process.env.NODE_ENV === 'undefined' && !debug) {
     process.env.NODE_ENV = 'production'
   }
   process.env.REACT_STATIC_ENV = 'production'


tannerlinsley:
Made a comment in #500. Thoughts?

denis-sokolov:
You are right that this changes the code from the final build, but so does not minifying it. Sure, `NODE_ENV!=production` can change behavior of ill-authored libraries, but so can minification, albeit less likely. On a scale from being identical to the production build and being very different, today the `build --debug` is one step away from the production build. Removing `NODE_ENV=production` would take it one more step away.

It is a trade-off: the build is less like production, but it’s more useful. There already exists a build that is exactly like production: the production build. There should be a clear value proposition for this different build, and library warnings seem like that.

tannerlinsley:
Excellent point. I'm on board now :)

tannerlinsley:
Are there any docs to update with this?

denis-sokolov:
No, I don’t think we need to explain it in the docs so far.

PR #513, opened 2018-03-23
Add Browser Support link to navigation. Detail how to extend static.c…

thislogancall:
Add Browser Support link to navigation. Detail how to extend static.confg.js to enable IE support through babel/promise polyfill

 ### Is this a bug report?
 
No


tannerlinsley:
Nice work!

PR #520, opened 2018-03-26
Fix api docs: getConfig noindex

pdjota:
Change from `noIndex` to `noindex`.
Fix indent on `dev: Boolean`.

Related warning when running `build`:
https://github.com/nozzle/react-static/blob/master/src/static/getConfig.js#L122I

 
 ### Is this a bug report?
 No. A small doc fix.


tannerlinsley:
Good find!

PR #526, opened 2018-03-28
Fix 404s on homepage at first load

### LINKED ISSUE #523: 404 error on home page
This errors occurs since **5.6.3** (I cannot reproduce for **5.6.2**). I haven't tested in production, so I can only guarantee it happens in development. 

Relevant static config :

``` javascript
getRoutes: () => [
    {
      path: '/',
      component: 'src/pages/Home',
    },
]
```

So I'll stick to 5.6.2 waiting for a fix ! If you need a reproduction I might setup one, but please have a close look at 5.6.3 changes before 44b390a2b4420d8cdd633cc2ccf584a473090458 commit since it should be hidden somewhere here.

mihai-tatar:
When first loading an app in dev mode there was a request to `/_react-static_/routeinfo/undefined`. This fixes the issue.

fixes #523

### DIFF
--- a/src/client/components/Router.js
+++ b/src/client/components/Router.js
@@ -67,7 +67,7 @@ export default class Router extends React.Component {
     (async () => {
       if (typeof window !== 'undefined') {
         // Get the entry path from location
-        const href = decodeURIComponent(window.location)
+        const href = decodeURIComponent(window.location.href)
         const path = cleanPath(href)
 
         // Injest and cache the embedded routeInfo in the page if possible


mihai-tatar:
@tannerlinsley, I saw you fixed something similar in 89d899dbc3b11d7ee600130a02c11ba298d5e7d2 so not sure if this PR is still mandated.

tannerlinsley:
Technically, this should be needed 🤔 . But why is it working right now then? Or is it not...

tannerlinsley:
WTF...
screen shot 2018-03-28 at 1 00 10 pm
screen shot 2018-03-28 at 1 00 16 pm

tannerlinsley:
Apparently, `window.location.toString()` returns a string similar to the `window.location.href`, So it works, but we should probably use `href` to make sure we're backwards compatible.

PR #540, opened 2018-04-04
Update README.md Sites Built with React-Static: NYC Vintage Map

sampok:
New site built with React Static: [NYC Vintage Map](https://nycvintagemap.com)


tannerlinsley:
Awesome!

PR #554, opened 2018-04-11
Gentics Mesh CMS Example

rafacm:
Template example based on the React Static [basic example](examples/basic) and the [Gentics Mesh CMS](https://getmesh.io/) [Angular Example](https://github.com/gentics/mesh-angular-example).


rafacm:
> Ready to merge?
@tannerlinsley: yes!

I will be improving the example in the coming weeks and fixing some of the issues mentioned in the README but IMHO it already is good enough.

> "Perfection is the enemy of progress" 
> -- Winston Churchill

PR #555, opened 2018-04-11
Do not scroll-to-top on initial page load

### LINKED ISSUE #551: Undesirable scroll-to-top on page refresh
First of all, thanks for React-Static – it’s great 👏

There’s one undesirable effect, though:

If I scroll down the page and refresh it, after the refresh the page scrolls back to the top with an animation (this can be seen on [react-static.js.org](https://react-static.js.org/)). And I expect scroll position to stay exactly where it was before the refresh – that’s how it behaves on browsers out of the box.

I’m aware of [props that are available on `Router` component](https://github.com/nozzle/react-static/blob/5fe177934381da6a79e15d4c5d96ef5d77a6a35d/docs/components.md#router), but I was still not able to work around this behaviour:

- If I set `autoScrollToTop={false}`, I lose auto scroll-to-top when the site path changes. Interestingly, it does not fix the scroll-to-top on page refresh, which still happens with this prop off.
- If I set `autoScrollToHash={false}`, the issue disappears, but then I lose animated scroll to hash, which some of my pages use. So it seems that scroll to hash is triggered on page refresh and that makes the page scroll back to the top.

Do you agree that this should be fixed?

Cinamonas:
This fixes the undesirable behaviour described in #551.

### DIFF
--- a/src/client/components/RouterScroller.js
+++ b/src/client/components/RouterScroller.js
@@ -7,7 +7,8 @@ import scrollTo from '../../utils/scrollTo'
 
 const RouterScroller = withRouter(class RouterScroller extends React.Component {
   componentDidMount () {
-    this.scrollToHash()
+    // Do not scroll to top on initial page load if hash does not exist
+    this.scrollToHash({ orScrollToTop: false })
   }
   componentDidUpdate (prev) {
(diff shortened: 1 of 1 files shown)


tannerlinsley:
Excellent!

PR #556, opened 2018-04-12
Fixes issue with error page showing in production

### LINKED ISSUE #553: `Router` component catches and shows (to users/clients) errors in production
The error print out appears to ship with the production build of `react-static`.
screen shot 2018-04-10 at 14 57 02

Also as `react-static` is catching the errors they are not getting to the top of our app where caught errors in production are sent to Sentry (https://sentry.io/welcome/)

We have fixed this by catching errors before they bubble to the `react-static` `<Router/>` component, but would prefer to change the behaviour of `react-static`.

Version: v5.6.7
https://github.com/nozzle/react-static/blob/v5.6.7/src/client/components/Router.js

### Suggested solution
`ErrorCatch` component that in development wraps the `<ResolvedRouter />` with a component that has `componentDidCatch` and prints out the UI. And in production it just returns `this.props.children`, so it doesn't have component with a `componentDidCatch` method and the error can bubble to the `<App />`/top level component. 

If people want to maintain the existing behaviour, errors shown in production, maybe a prop on the `<ResolvedRouter />` component  like `showErrorsInProduction={true}` could be an option.

jasonlafferty:
This pull request decouples error handling and UI  from the `<Router />` and moves it into the `<ErrorWrapper />` . Introduces additional prop on the `<Router />`, `showErrorsInProduction` which when set to `true` to enable the debug/error page in production. Defaults to `false`, error page does not appear in production.

This pull request fixes #553 and introduces some tests arounds the changes.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,3 +1,6 @@
 module.exports = {
   extends: 'react-tools',
+  env: {
+    jest: true,
+  },
 }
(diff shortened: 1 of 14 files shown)


tannerlinsley:
Excellent! Nice work with the tests too!

PR #572, opened 2018-04-19
Added new configuration option 'disableRoutePrefixing'

rachel-church:
## Description
Disabling the route prefix will allow the usage of a base URL that will apply to most resource GET requests but not history and route URL requests.

An example use case would be `mysite.com/us/en/blog` which would request for `/us/en/blog/index.html` and `/us/en/blog/routeInfo.json`. But allowing other root resources to be at a different baseUrl such as `my-blog`. Loading the `mysite.com/us/en/blog` would request for `/my-blog/bootstrap.981e38c5.js`

## Changes/Tasks
- [ ] Added new boolean setting `config.disableRoutePrefixing`

## Motivation and Context
https://github.com/nozzle/react-static/issues/571

## Types of changes
- [ x ] New feature (non-breaking change which adds functionality)

## Checklist:
- [ x ] My change requires a change to the documentation.
  - [ x ] I have updated the documentation accordingly.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -23,6 +23,7 @@ A `static.config.js` file is optional, but recommended at your project root to u
 * [outputFileRate](#outputfilerate)
 * [prefetchRate](#prefetchrate)
 * [disableRouteInfoWarning](#disableRouteInfoWarning)
+* [disableRoutePrefixing](#disableRoutePrefixing)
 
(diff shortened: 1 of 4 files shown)


tannerlinsley:
Nice work! I'll cut a release soon

PR #574, opened 2018-04-19
Don't extract CSS in webpack during node stage

lottamus:
Currently in 5.8.1, CSS files are getting extracted in both node and prod stages. This results in two identical CSS files.

This PR removes `ExtractCssChunks` and  `ExtractTextPlugin` when in the node stage.

## How to reproduce

Run `yarn run text-sample-build` and notice it generates two CSS files with identical content.

image

image

image


tannerlinsley:
An excellent find! Great work.

PR #598, opened 2018-05-04
Fix examples that override webpack css-loaders

### LINKED ISSUE #588: Sass example cannot be built
### Environment

1.  `react-static -v`: 5.8.2+
2.  `node -v`: n.a.
3.  `yarn --version or npm -v`: n.a.
4.  Operating system: n.a.

**Base your steps off of any freshly installed react-static template!**

1. `cd examples/sass`
2. `npm run build`

### Expected Behavior

A successfull build

### Actual behaviour

```
ERROR in <path to>.scss
Module build failed: Error: "extract-text-webpack-plugin" loader is used without the corresponding plugin, refer to https://github.com/webpack/extract-text-webpack-plugin for the usage example
    at Object.pitch ( <path to project>/node_modules/extract-text-webpack-plugin/dist/loader.js:57:11)
 @ ./dist/react-static-routes.js
 @ ./src/App.tsx
 @ ./src/index.tsx


```

lottamus:
fixes #588

## Description

This is the same issue that was fixed with https://github.com/nozzle/react-static/pull/574 where we don't want to extract CSS files in both "prod" and "node" stages. This would result in two duplicate CSS files.

During the node stage we don't include the ExtractTextPlugin. This example was breaking because it was trying to use that plugin during the node stage.

## Changes/Tasks

- [x] Only extract css to a file during prod stage
- [x] update package-lock.json since it was trying to use react-static 4.3.4

## Types of changes

### DIFF
--- a/examples/less-antdesign/package.json
+++ b/examples/less-antdesign/package.json
@@ -10,7 +10,7 @@
     "serve": "serve dist -p 3000"
   },
   "dependencies": {
-    "antd": "3.3.0",
+    "antd": "3.5.0",
     "axios": "^0.18.0",
     "react": "^16.2.0",
     "react-dom": "^16.2.0",
--- a/examples/less-antdesign/src/App.tsx
+++ b/examples/less-antdesign/src/App.tsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import { Link, Router, withRouter } from 'react-static'
+import { Link, Router } from 'react-static'
 import styled, { injectGlobal } from 'styled-components'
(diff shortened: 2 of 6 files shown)


Reggino:
Thanks! 

The same change applies to the "tailwind"-example (and others) i think...

Reggino:
I think the `less-antdesign`-example is the last one...

Helpothers2day:
Looking forward to working with you all!!😃

Helpothers2day:
Yes I love Learning

PR #603, opened 2018-05-11
Adds Automated building of all examples

sudarshang:
Adds a script to build all the examples in an automated manner.

## Description
testBuildsExample.js is a script that builds all the examples.

## Changes/Tasks
- [x] Added code

## Motivation and Context
It is possible that changes to the react-static library can cause some of the examples that are shipped with the library to stop working. There is no easy way for a developer to check if their changes are breaking any of the templates that react-static supports. This script builds every example and if any example fails to build it prints out the error and exits with an error code.

TODO:
- integrate this with travis

## Types of changes
- [x]  Added new test
## Checklist:

### DIFF
--- a/testExamplesBuild.js
+++ b/testExamplesBuild.js
@@ -1,0 +1,34 @@
+const { execSync } = require('child_process')
+const fs = require('fs')
+const chalk = require('chalk')
+
+const exampleToError = new Map()
+
+fs.readdirSync('./examples').forEach(
+  file => {
+    process.chdir(`./examples/${file}`)
+    try {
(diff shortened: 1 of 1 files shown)


sudarshang:
Running this script I find that following examples are failing on master

- firebase-auth
- gentics-mesh
- less-antdesign
- material-ui
- sass
- typescript

See this [gist](https://gist.github.com/sudarshang/5104bf10e6404a1b3d0a8191dca2a75c) for detailes

tannerlinsley:
Looks like we need to check on those examples!

PR #608, opened 2018-05-14
fix(sitemap): rm trailing slash

### LINKED ISSUE #354: Sitemap slashes duplicated
Sitemap location url contains duplicate slashes, regardles of having trailing or no siteRoot slash.

Version in use 5.1.6.
`siteRoot: 'https://site.com'`

Expected:
`<loc>https://site.com/route/</loc>`

Result:
`<loc>https://site.com//route/</loc>`

davidpfahler:
Fixes #354.

## Changes/Tasks

- [x] Changed code

## Types of changes

## Checklist:

### DIFF
--- a/src/static/index.js
+++ b/src/static/index.js
@@ -16,7 +16,7 @@ import jsesc from 'jsesc'
 //
 import generateRoutes from './generateRoutes'
 import { DefaultDocument } from './RootComponents'
-import { poolAll, pathJoin, trimLeadingSlashes } from '../utils/shared'
+import { poolAll, pathJoin, trimLeadingSlashes, trimTrailingSlashes } from '../utils/shared'
 import Redirect from '../client/components/Redirect'
 
 const defaultOutputFileRate = 100
@@ -455,7 +455,7 @@ export async function buildXMLandRSS ({ config }) {
   const prefixPath = config.disableRoutePrefixing ? siteRoot : config.publicPath
   const xml = generateXML({
     routes: config.routes.filter(d => !d.is404).map(route => ({
-      permalink: `${prefixPath}/${trimLeadingSlashes(pathJoin(route.path))}`,
+      permalink: `${trimTrailingSlashes(prefixPath)}/${trimLeadingSlashes(pathJoin(route.path))}`,
       lastModified: '',
       priority: 0.5,
       ...route,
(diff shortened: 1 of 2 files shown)


src/static/index.js
```diff
@@ -452,7 +452,8 @@ export async function buildXMLandRSS ({ config }) {
   if (!siteRoot) {
     return
   }
-  const prefixPath = config.disableRoutePrefixing ? siteRoot : config.publicPath
+  let prefixPath = config.disableRoutePrefixing ? siteRoot : config.publicPath
+  prefixPath = prefixPath.replace(/\/$/, '')
   const xml = generateXML({
     routes: config.routes.filter(d => !d.is404).map(route => ({
       permalink: `${prefixPath}/${pathJoin(route.path)}`,
```
lottamus: You could still get a double slash here if `route.path === '/'`. Here is the code for `pathJoin`: https://github.com/nozzle/react-static/blob/master/src/utils/shared.js#L6

Perhaps `trimSlashes` should be broken into two functions: `trimLeadingSlashes` and `trimTrailingSlashes`. Then we can use those everywhere?

https://github.com/nozzle/react-static/blob/master/src/utils/shared.js#L52

tannerlinsley: Yeah, there are so many edge cases with slashes everywhere. I would much rather have all of them going through utility functions that have tests.

anjawaldbauer: Makes sense. Will update.

src/static/index.js
```diff
@@ -455,7 +455,7 @@ export async function buildXMLandRSS ({ config }) {
   const prefixPath = config.disableRoutePrefixing ? siteRoot : config.publicPath
   const xml = generateXML({
     routes: config.routes.filter(d => !d.is404).map(route => ({
-      permalink: `${prefixPath}/${pathJoin(route.path)}`,
+      permalink: `${prefixPath}/${trimLeadingSlashes(pathJoin(route.path))}`,
```
lottamus: Still runs into the same issue if prefixPath ends in a slash. 

Needs 

```
permalink: `${trimTrailingSlashes(prefixPath)}/${trimLeadingSlashes(pathJoin(route.path))}`,
```

davidpfahler:
@lottamus Thanks. Let me know if this is better

davidpfahler:
Closing as this seems to be fixed in https://github.com/nozzle/react-static/commit/597f18f46a0d20d6663bc782b97f4641be7a35ed

PR #619, opened 2018-05-22
Report errors when building configuration

inyono:
## Description

Adds logging of compiler errors during requiring of config file (like `static.config.js`)

## Changes/Tasks

- [x] Added `console.error` statement to `catch` block of `buildConfigFromPath`

## Motivation and Context

When `static.config.js` contains any compiler errors causing `require(configPath).default` to fail, `react-static build` fails with a rather unhelpful error message (since it tries to continues with an empty configuration)
``` TypeError: Cannot read property 'DIST' of undefined```
This makes debugging errors in `static.config.js` hard.

## Types of changes

## Checklist:

### DIFF
--- a/src/static/getConfig.js
+++ b/src/static/getConfig.js
@@ -198,6 +198,7 @@ const buildConfigFromPath = configPath => {
     const config = require(configPath).default
     return buildConfigation(config)
   } catch (err) {
+    console.error(err)
     return {}
   }
 }


sudarshang:
Thanks! I was bitten by this one yesterday. Should we use `chalk` to ensure that the error is more visible?

inyono:
Another question: does it even make sense to try to continue with an empty config? Or is there some config merging somewhere else that needs this? With this change, both the compiler error and the weird error are shown (both with stacktrace). The second error doesn't help non-contributors though, I suppose.

tannerlinsley:
I'm okay with this for now. It may be superfluous, but I'd rather have more visibility than not for now.

PR #632, opened 2018-06-10
fix for issue#631

marsinearth:
## Description

I've figured out the workaround to make `yarn build` work in react-static@5.9.7 with typescript, otherwise it returns: 
```
Module build failed: Error: "extract-text-webpack-plugin" loader is used without the corresponding plugin, refer to https://github.com/webpack/extract-text-webpack-plugin for the usage example
```

## Changes/Tasks

I referred briefly on [extract-text-webpack-plugin example](https://github.com/webpack-contrib/extract-text-webpack-plugin#usage), added a few lines. You can easily figure out what's going on at a glance I believe.

## Motivation and Context

It's an [open issue](https://github.com/nozzle/react-static/issues/631)

## Types of changes

## Checklist:

### DIFF
--- a/examples/typescript/static.config.js
+++ b/examples/typescript/static.config.js
@@ -83,7 +83,7 @@ export default {
         ],
       },
     ]
-    // You might need to make the file name universal
+    // FIXME: You might need to make the file name universal
     config.plugins = [new ExtractTextPlugin('index.css')]
     return config
   },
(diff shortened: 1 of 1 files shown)


examples/typescript/static.config.js
```diff
@@ -68,11 +71,20 @@ export default {
               },
             ],
           },
-          defaultLoaders.cssLoader,
+          {
+            test: /\.css$/,
+            use: ExtractTextPlugin.extract({
```
inyono: `defaultLoaders.cssLoader` works for me, too

tannerlinsley:
This should no longer be necessary with the new webpack 4 upgrade.

xlwknx:
But it is still not working

PR #633, opened 2018-06-11
Sitemap.xml may contain invalid characters

Reggino:
I had some issues with Google rejecting my sitemap.xml because it contained invalid XML. Some XML characters need te be escaped while generating sitemap.xml (e.g. `&`)

This should fix that.


Reggino:
Not sure why travis build is failing, but i don't think it is related to the change...

PR #636, opened 2018-06-14
W3C HTML validator warns about HTML

Reggino:
W3C HTML validator warns about superfluous / deprecated use of `type="text/javascript"` in <script>-tags in  HTML 5.

This change removes them.


Reggino:
Hhmm.. injected script tags come from one of the node_modules. 

Will work on a better fix.

PR #637, opened 2018-06-14
Added Webshop / E-commerce use to the example sites

Reggino:
The content is in Dutch, so probably not that relevant. But i may be nice to see react-static used in a different language, in a webshop :-)


Rusfighter:
@Reggino Hey Tim, I am also dutch and I really like the shop you made. Currently I am also busy creating my own dropshipping ecommerce store and I wonder what headless cms you used for the inventory (product info/images). 

Perhaps we can talk in private since this is off-topic

Reggino:
@Rusfighter Thanks, but I'm afraid i can't help you there. The back-end is fully custom made and closed source. But any e-commerce solution with some sort of API should suffice.

PR #659, opened 2018-07-10
Fix issue with lots of routes where config cannot be stringified

bkniffler:
Fix issue with lots of routes where config cannot be stringified.
Fixes https://github.com/nozzle/react-static/issues/658


tannerlinsley:
Nice find!

PR #677, opened 2018-07-15
[FIX]: Fixing version mismatch in lockfile

ScriptedAlchemy:
## Description
Locks were causing my extract-css-chunks plugin to fail, installing v2 (webpack 3) and not v4 (webpack 4)

## Motivation and Context
Well, stuff dont work otherwise :) 

## Types of changes


ScriptedAlchemy:
@tannerlinsley here ya go :)

PR #678, opened 2018-07-15
Fix single thread exporting

KidkArolis:
## Description
Fix `maxThreads: 1` case.

## Changes/Tasks
The exporter now takes `routes` as a seperate arg, separate from config.

## Screenshots (if appropriate):
image

## Types of changes


tannerlinsley:
I think you'll need to make a change here too https://github.com/nozzle/react-static/blob/master/src/static/index.js#L205

KidkArolis:
Hm, I did?

image

Or did you mean something else?

PR #689, opened 2018-07-21
[Bugfix] Remove path="*" since its broken with latest dependencies

bkniffler:
react-router uses path-to-regexp, which will not recognize `*`

Read
https://github.com/pillarjs/path-to-regexp#compatibility-with-express--4x
=> `No wildcard asterisk (*) - use parameters instead ((.*))`

But leaving the path away in react-router `Route` will effectively match all routes.

Please publish if you accept this change :) I'm living on the edge!

### DIFF
--- a/src/static/generateRoutes.js
+++ b/src/static/generateRoutes.js
@@ -155,7 +155,7 @@ export default class Routes extends Component {
 
     // This is the default auto-routing renderer
     return (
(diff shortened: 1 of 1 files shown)


bkniffler:
Heya @tannerlinsley, could you publish this change? Thanks!

PR #719, opened 2018-08-28
update to babel7.0 and other deps updates

digitalkaoz:
EOM


bkniffler:
If this could be merged and published, I'd be very happy!

nozzle:
Info: `master` branch now points to 5.x.x and the `v6` branch to 6.0.0-beta. You may need to close and reopen your PR to the v6 branch.

package.json
```diff
@@ -114,16 +114,16 @@
     "enzyme": "^3.4.3",
     "enzyme-adapter-react-16": "^1.2.0",
     "enzyme-to-json": "^3.3.4",
-    "eslint": "4",
-    "eslint-config-prettier": "^2.9.0",
+    "eslint": "^5.4.0",
```
tannerlinsley: Last I checked, this version of eslint still had issues. I think we should hold off on this one for a bit more... Thoughts?

digitalkaoz: i expected to issues. do you have any links?

tannerlinsley:
No but I remember there being an incompatibility with the Airbnb config.

digitalkaoz:
worked correctly while crafting this PR, so i dont know

tannerlinsley:
We'll go with it then. If it pops up again, we can handle it easily.

PR #724, opened 2018-09-01
Make message port configurable

mikekelly:
The message port is hard-coded to 4000, this makes it configurable


nozzle:
Info: `master` branch now points to 5.x.x and the `v6` branch to 6.0.0-beta. You may need to close and reopen your PR to the v6 branch.

PR #746, opened 2018-09-12
[CHORE]: Updating RUC (react-universal-component)

ScriptedAlchemy:
[CHORE]: Updating RUC (react-universal-component)

Should get rid of those "critical dependency is a expression" warnings you see in webpack 4.16

<3 <3


tannerlinsley:
Excellent!

PR #748, opened 2018-09-12
Fix hot reloading when modifying pages from routes.

lucasmogari:
Using react-static v5.9.8, hot reloading only works on App.js. Adding the line below in .babelrc fix it. I only tested on theses examples, but this problem might happen in others.

plugins": ["react-hot-loader/babel"]

See https://github.com/nozzle/react-static/issues/725


tannerlinsley:
This is very strange, because the react-static/babel-preset.js has the plugin included already.

lucasmogari:
I know this is not a fix for the real problem, but I think the examples should just work after downloading for people that are testing react-static. Feel free to ignore and close it. No hard feelings lol

tannerlinsley:
This is v5 and will soon be outdated, so it's not a huge priority. But at the same time, I want people who use it until v6 to be happy.

I will merge this PR if you add the changes to all the remaining example directories. Thoughts?

lucasmogari:
I can do that.

tannerlinsley:
Nice! Thanks 😄

PR #751, opened 2018-09-14
bump netlifycms version to 2.1

sw-yx:
Hi Tanner! just tried the netlifycms template out and it failed when i was authenticated. did some digging and found you were still using 1.0 and the problem went away when i bumped the version to 2.1 (latest). so here's a one line PR.

## Changes/Tasks
- [x] bumped one line in the html

## Motivation and Context
netlifycms template doesnt work anymore without this upgrade

## Types of changes

## Checklist:


tannerlinsley:
Perfect!

tannerlinsley:
In v6, I hope to have netlify cms be a plugin you just add

PR #758, opened 2018-09-20
[v6] Don't recompile DIST dir when exporting HTML

lottamus:
## Description

> `@babel/register` uses Node's require() hook system to compile files on the fly when they are loaded. 

This means that https://github.com/nozzle/react-static/blob/v6/src/static/exporter.js#L16 and https://github.com/nozzle/react-static/blob/v6/src/static/threadedExporter.js#L19 will require a file that has already been compiled and minified. This file can be fairly large (1mb+) which means babel might run out of memory when trying to parse and compile again.

## Changes/Tasks

- Added a way to set an ignore path during runtime

## Motivation and Context

Fixes https://github.com/nozzle/react-static/issues/757

## Screenshots (if appropriate):

Fixes this error:

image

## Types of changes

## Checklist:


lottamus:
Wrong branch

PR #768, opened 2018-09-26
remove possible leading path separator

digitalkaoz:
fixes an case in my previously introduced feature


tannerlinsley:
Nice find :)

PR #769, opened 2018-09-26
fix double compiling of node_modules

digitalkaoz:
## Description

https://github.com/nozzle/react-static/commit/ad87562fd912acf2479c2843937ceee8cc7e3673 introduced some double compiling.
like here: https://github.com/glimmerjs/glimmer-application-pipeline/issues/87

so excluding `node_modules` we must!

## Changes/Tasks
- [x ] Changed code

## Types of changes

## Checklist:

### DIFF
--- a/src/static/webpack/rules/jsLoader.js
+++ b/src/static/webpack/rules/jsLoader.js
@@ -37,6 +37,7 @@ export default function({ config, stage }) {
         loader: 'babel-loader',
         options: {
           ...babelFile,
+          exclude: new RegExp(`(node_modules|${config.paths.EXCLUDE_MODULES})`),
           root: config.paths.ROOT,
           presets: [[babelPreset, { modules: false }]],
           cacheDirectory: isRelativePath ? stage !== 'prod' : config.paths.TEMP,


tannerlinsley:
I was having trouble with an exclude here. It was killing just about every external import from node_modules (emotion, react-table, etc) from properly importing (trying to import module.default.default haha).

Have you tested this with any of those?

Also, I removed the exclude mostly based on what CRA had done by simply deciding to compile everything from node_modules. It ended up being the only way they could allow people to use ESM and commonjs modules in the same projects.

Thoughts?

digitalkaoz:
no i didnt tested it with ESM packages. but without excluding it gets even worse here:

`_typeof is not a function` in the webpack bundle, and nothing works. i dont know which package caused this problem. i guess its happening because of the double transpile of some already babel transpiled code. which makes it even worse i think.

are there ESM only packages out there? if yes we should find some other way like this:

> exclude `node_modules` but not `node_modules/react-table` and not `node_modules/emotion` and so on

or we leave the exclude option, but not excluding `node_modules` by default? so one can exclude if it have to?

or leave it up to the local babel config? so many questions.

tannerlinsley:
For reference, this is where CRA double compiles to avoid this problem: https://github.com/facebook/create-react-app/blob/next/packages/react-scripts/config/webpack.config.prod.js#L313-L353

digitalkaoz:
@tannerlinsley if #775 works we can close this one...

PR #773, opened 2018-09-28
Add .nvmrc

ajschmidt8:
Any objections to adding a `.nvmrc` file? Makes things easier for contributors who might not use LTS by default.


tannerlinsley:
Good thinking ;)

JustFly1984:
I'm using node 10.11 in production, staging and development and I have my own .nvmrc file. Please make sure that I can overwrite react-static's file in the root.

PR #775, opened 2018-09-28
compile node_modules with bare minimum too

digitalkaoz:
as a general exclude on node_modules doesnt seem to be a good idea, we compile them too (but with a bare minimum of features). the idea is inspired by CRA https://github.com/facebook/create-react-app/blob/next/packages/react-scripts/config/webpack.config.prod.js#L313-L353

## Description

separate babel loaders for SRC and NODE_MODULES

## Changes/Tasks
- [x] Changed code

## Motivation and Context
to be able to use external ESM 

## Types of changes

## Checklist:

### DIFF
--- a/src/static/webpack/rules/jsLoader.js
+++ b/src/static/webpack/rules/jsLoader.js
@@ -32,7 +32,9 @@ export default function({ config, stage }) {
 
   return {
     test: /\.(js|jsx|mjs)$/,
+    include: [config.paths.SRC, `${config.paths.DIST}/react-static-routes.js`],
     use: [
+      // 'thread-loader',
       {
         loader: 'babel-loader',
         options: {
--- a/babel-preset-external.js
+++ b/babel-preset-external.js
@@ -1,0 +1,74 @@
+const r = require.resolve
+
+module.exports = (api, options = {}) => {
+  const { NODE_ENV, BABEL_ENV } = process.env
+
+  const PRODUCTION = (BABEL_ENV || NODE_ENV) === 'production'
+  const DEVELOPMENT = (BABEL_ENV || NODE_ENV) === 'development'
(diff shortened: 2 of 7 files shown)


digitalkaoz:
@tannerlinsley another attempt of #769 maybe you can give it a try with some external ESM packages too?! it seems to work here (https://github.com/uber/react-map-gl/pull/599 is an outstanding issue, but probably not related)

src/static/webpack/rules/jsLoader.js
```diff
@@ -32,7 +32,9 @@ export default function({ config, stage }) {
 
   return {
     test: /\.(js|jsx|mjs)$/,
+    include: [config.paths.SRC, `${config.paths.DIST}/react-static-routes.js`],
     use: [
+      // 'thread-loader',
```
tannerlinsley: Does `thread-loader` not work properly here?

digitalkaoz: it threw some error that i didnt found:

```
Module build failed (from node_modules/thread-loader/dist/cjs.js):
Thread Loader (Worker 1)
.presets[0][0] must be a string, object, function
    at Array.forEach (<anonymous>)
    at Array.forEach (<anonymous>)
```

but it is:

```js
      {
        loader: 'babel-loader',
        options: {
          babelrc: false,
          configFile: false,
          compact: false,
          presets: [
            [
              babelPreset,
              { helpers: true },
            ],
          ],
          cacheDirectory: true,
          sourceMaps: false,
        },
      },
```

so im not sure where the problem is. the docs stated anyway that its only useful for large projects (starting a worker gives 200ms overhead), so if we want to use it, it should be some optin strategy i think

tannerlinsley:
Sounds good. I tested this and it seems to work with the libraries I had in question so far.

PR #794, opened 2018-10-05
make the static files location more flexible

digitalkaoz:
somehow the `tmp` paths on OSX behave weird, so we simple change the static files to a regex on the filename


digitalkaoz:
the CI seems to be timed out...

Issue #649, opened 2018-06-29
Intermittent dynamic route error


cdock1029:
### Are you reporting a bug?
Yes

I have a single non-static `/admin` route. 

[Like this commenter on Spectrum describes](https://spectrum.chat/?t=35024ee8-3a03-48db-9ea0-d6c88cf539e9), I'm seeing an error (*sometimes*) when loading the admin route:

```js
Uncaught (in promise) TypeError: Cannot set property 'templateLoaded' of undefined
```

I can navigate successfully on client, and load page from server. Every so often upon a reload of the app, then navigating to `/admin` I see that error. 

The Spectrum comment is addressing something else with more complex dynamic routes. But mine is a single route, as per the non-static route example, showing the same error. 

Caching bug? Race condition? 

Location of error: 

https://github.com/nozzle/react-static/blob/63e2eb69ffc47263cf3e6de504d9778063f7ae29/src/client/methods.js#L209

### Environment

1.  `react-static -v`: `5.9.7`
2.  `node -v`: `v9.11.1`
3.  `yarn --version or npm -v`: `npm: 6.1.0`
4.  Operating system: `macOS Sierra 10.12.6`
5.  Browser and version (if relevant): `Chrome 69.0.3472.3`

### Steps to Reproduce the problem

**Base your steps off of any freshly installed react-static template!**

1.  Load app, any route (/admin should work)
2.  Navigate around, works
3.  Reload app... sometimes client navigation to /admin produces error above.

`static.config.js`
```js
getRoutes: async () => {
    const value = await admin
      .database()
      .ref('health')
      .once('value')
      .then(snap => snap.val())
    // const snap = await snapPromise
    console.log({ value })
    const membersLoad = loadEntity('members')
    const eventsLoad = loadEntity('events')
    const linksLoad = loadEntity('links')

    const [members, events, links] = await Promise.all([
      membersLoad,
      eventsLoad,
      linksLoad,
    ])

    return [
      {
        path: '/',
        component: 'src/containers/Events',
        getData: () => ({ events }),
      },
      {
        path: '/members',
        component: 'src/containers/Members',
        getData: () => ({ members }),
      },
      {
        path: '/links',
        component: 'src/containers/Links',
        getData: () => ({ links }),
      },
      {
        is404: true,
        component: 'src/containers/404',
      },
    ]
  },
```

`App.js`
```jsx
const Admin = universal(() => import('../containers/Admin'))
const App = () => (
  <Router>
    <div id="app">
[...]

PR #813, opened 2018-10-12
Fix templateLoaded bug/error that occurs in non-static routes

boonsuen:
## Description

There's a bug with non-static routes. See #649. When you landed (refresh or visit directly) on a non-static routes, suppose it is /admin, whether you're in the same route or a different route if you navigated away, you cannot navigate back to the /admin route. You will get the `Cannot set property 'templateLoaded' of undefined` error. 
I get it to work by only update the templateLoaded property when routeInfo is indeed available.

## Changes/Tasks

- [x] Changed code

## Types of changes

## Checklist:


tannerlinsley:
Update this PR to *not* include the built `/lib` directory, then we'll take a look at the changes.

boonsuen:
Open up a new PR with another source branch.

PR #819, opened 2018-10-15
Go Router-less!

tannerlinsley:
This branch removes most of the core necessity of react-router and a router in general.

This leaves the user to implement whatever `Link`ing approachthey want, so long as it uses `history.push/replaceState` under the hood (every react router does this ;)

We are going to still recommend people use Reach Router, which will be the router we use in all of our examples as well. (We need to migrate away from React-Router for all of the examples.

From the changelog:
- All `render` and `component` props are now deprecated in favor of using `child-as-a-function` rendering. Anywhere you are using these props must be migrated to use a child as a function.
- The `Loading` component has been removed. If you wish to show loading states in your app, you can use the `Loading` props for any components that support them. This is in preparation for React...Suspense!
- React-Static no longer ships with any routing-related functionality. It functions independently of your routing paradigm. Thus, it no longer exports anything from `react-router`.
- The `PrefetchWhenSeen` component has been deprecated in favor of only using the `Prefetch` component
- The `Prefetch` component is now smart like `PrefetchWhenSeen` was.
- The client-side `Redirect` component has been deprecated. Redirects should be done in the `static.config.js`. If the user needs to do any redirects for dynamic/runtime routes, they can use their favorite router's redirect solution.
- `Router` has been deprecated and replaced by the `Root` component. The `Root` component implements the `HashScroller` component, an `ErrorBoundary` and a very simple and non-invasive route context using `@reach/router` (the recommended router). The base router is customizable or replaceable if the user wishes to use a different router.


tannerlinsley:
It should be noted, I'm still working on getting this branch to build correctly, but the gist of it is there. Not much more is going to change... tentatively I guess.

tannerlinsley:
Re: Router 

Totally agree. I am planning on documentation and all of the examples demonstrating and shipping with Reach Router with productivity as a priority. For people getting started, they shouldn't have to worry about a router. And for people who want to go crazy with dynamic routing solutions, they can quite literally go crazy by replacing the `Root` components render function with their own router.

99% use-case will be opinionated towards reach/router though.

bkniffler:
I just tried running the example.
- Automatic routes from `pages` folder are not picked up resulting in an error due to index route not found. This is due to getRoutesFromPages.js 
```jsx
const handle = pages => {
    // Turn each page into a route
    const routes = pages.map(page => {
      // Get the component path relative to ROOT
      const component = nodePath.relative(config.paths.ROOT, page)
      // Make sure the path is relative to the root of the site
      let path = page.replace(`${config.paths.PAGES}`, '').replace(/\..*/, '')
      // Turn `/index` paths into roots`
      path = path.replace(/\/index$/, '/')
      // Return the route
      return {
        path,
        component,
        isPage: true, // tag it with isPage, so we know its origin
      }
    })
    return routes
  }
```
This only works for unix paths, not windows I guess. The component value is relative (`src\\pages\\404.js`) and the path value is absolute (`C:/Users/bkniffler/GitHub/project/src/pages/404`).

- Plugins don't work at all. Tried adding typescript like in the example. The plugin is not found if it is in `plugins/name-of-plugin.js` and with `plugins: ['name-of-plugin']`. It is found if stored as `plugins/name-of-plugin/index.js`. It is not being called either way though.

karlmacklin:
Will you start reverting back to react-router instead of reach-router from now on, seeing as reach-router will be abandoned eventually?

Currently going through hell trying to introduce react-static to our webapp with newer react-router. No idea how to setup correctly at the moment (how to get the history object from react-static?). Pushes to history does not reflect in react-router inside React components.

Issue #818, opened 2018-10-15
material-ui template is using react-jss, but not included in the package.json


Rantelo:
**Describe the bug**
when running `yarn start` after creating a site with `material-ui` template, it throws an error saying `Cannot find module 'react-jss/lib/jss`

**To Reproduce**
1. Create a new project with `react-static create`
2. Select `material-ui` template
3. Run dev-server with `yarn start`
```
$ react-static start
  Error: Cannot find module 'react-jss/lib/jss'
  
  - module.js:542 Function.Module._resolveFilename
    module.js:542:15
  
  - module.js:472 Function.Module._load
    module.js:472:25
  
  - module.js:585 Module.require
    module.js:585:17
  
  - module.js:11 require
    internal/module.js:11:18
  
  - static.config.js:3 Object.<anonymous>
    /home/imuser/Documents/site/personalblog/static.config.js:3:1
  
  - module.js:641 Module._compile
    module.js:641:30
  
  - node.js:144 loader
    [personalblog]/[babel-register]/lib/node.js:144:5
  
  - node.js:154 Object.require.extensions.(anonymous function) [as .js]
    [personalblog]/[babel-register]/lib/node.js:154:7
  
  - module.js:560 Module.load
    module.js:560:32
  
  - module.js:503 tryModuleLoad
    module.js:503:12

```

**Expected behavior**
It should start dev-server using the template

**Desktop (please complete the following information):**
 - OS: Ubuntu 16.04 LTS
 - Browser Chrome Version 68.0.3440.84 (Official Build) (64-bit)

**Additional context**
I don't think this issue is related to react-static by itself, because I tried other templates and they worked just fine.
I believe the issue is related with the template

Herohtar:
This can be resolved with a simple `yarn add react-jss` or `npm i react-jss`. Material UI removed their `react-jss` dependency in version `3.2.0` so it no longer gets pulled in automatically.

Rantelo:
> This can be resolved with a simple `yarn add react-jss` or `npm i react-jss`. Material UI removed their `react-jss` dependency in version `3.2.0` so it no longer gets pulled in automatically.

Great! It worked
Any reason to delete `react-jss` from the package.json?
If there is not, I can file a PR addition of the missing library

Issue #713, opened 2018-08-22
Error: Cannot find module '../lib/utils/binHelper'


gastonmorixe:
`../lib/utils/binHelper` shouldn't be `../src/utils/binHelper` ?

elassol:
Did you solve the issue?

gastonmorixe:
@elassol nop. abandoned the idea of using react-static sadly until it works with babel 7 out-of-the-box.

briaguya:
I encountered this issue as well:
I first tried: `yarn add react-static@next` and ran into #701 
Then I tried adding master using `yarn add https://github.com/nozzle/react-static`
I then cd'd into the node_modules/react-static dir and ran `yarn build` so the lib folder would show up and it started working.

benadamstyles:
The `src` folder doesn't exist, if you install using `npm` rather than `yarn`!

nozzle:
Info: `master` branch now points to 5.x.x and the `v6` branch to 6.0.0-beta.

hmedney:
I've upgraded a 5.x project using both `yarn add react-static@latest` and `yarn add https://github.com/nozzle/react-static` - been hitting:

```bash
Error: Cannot find module '../lib/utils/binHelper'
    at Function.Module._resolveFilename (module.js:547:15)
    at Function.Module._load (module.js:474:25)
    at Module.require (module.js:596:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/mnt/d/dev/project/ReactStatic/node_modules/react-static/bin/react-static-start:4:1)
    at Module._compile (module.js:652:30)
    at Object.Module._extensions..js (module.js:663:10)
    at Module.load (module.js:565:32)
    at tryModuleLoad (module.js:505:12)
    at Function.Module._load (module.js:497:3)
```

vision669:
same issue please fix

robbyemmert:
What versions of NPM are react-static known to work with? Also, any idea when this was changed? What is the latest known working version of react-static?

robbyemmert:
I found the issue. They're tagging beta builds with the 'latest' tag, which makes npm install the latest beta version by default (which happens to be broken at the moment).

Workaround to install the latest stable version: 
npm uninstall -g react-static
npm install -g react-static@v5.9.5

[...]

PR #820, opened 2018-10-15
Add missing dependency to material-ui template

Rantelo:
## Description
Material-UI removed their dependency to `react-jss` in `v3.2.0` so it no longer gets pulled in automatically.

## Changes/Tasks
I added `react-jss` dependency to material-ui's template

## Motivation and Context
This PR solves #818


Rantelo:
Failed build with Travis CI seems to be related with issue #713

PR #829, opened 2018-10-22
Use updated UglifyJsPlugin for create-react-app compatibility

cooperka:
## Motivation

Fixes `Unexpected token: punc ())` errors during build.

## Changes

1. Commit automatic 'integrity' info from yarn 1.10

2. Use updated UglifyJsPlugin for create-react-app compatibility
    - Fixes `Unexpected token: punc ())` errors during build
    - See discussions for details:
        - https://github.com/webpack-contrib/uglifyjs-webpack-plugin/issues/104#issuecomment-358646531
        - https://github.com/webpack/webpack/issues/5858#issuecomment-338430720
    - See also [UglifyjsWebpackPlugin docs](https://webpack.js.org/plugins/uglifyjs-webpack-plugin/)

## Types of changes

- [ ] Refactoring/add tests (refactoring or adding test which isn't a fix or add a feature)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [ ] My change requires a change to the documentation.
  - [ ] I have updated the documentation accordingly.
- [ ] My changes have tests around them

Note: I've verified this change on my own production app, but I haven't tested much beyond that. You can see the build logs after the change [here](https://travis-ci.org/cooperka/neuro-name/builds/444617571).

PR #830, opened 2018-10-22
Update yarn integrity + prettier

cooperka:
Per https://github.com/nozzle/react-static/pull/829#issuecomment-431940885, I went to apply the changes from #829 to branch v6, but it looks like the plugin is already configured [here](https://github.com/nozzle/react-static/blob/v6/packages/react-static/src/static/webpack/webpack.config.prod.js#L89). I applied the automatic yarn changes, at least :tropical_drink:


tannerlinsley:
Awesome :)

PR #837, opened 2018-10-25
Fix typings & add react-static/node typing

Perlmint:
## Description
Fix existing typings & add react-static/node typing

## Changes/Tasks
- [x] Add `react-static/node` typing
- [x] Fix wrong types in typing
- [x] Specify types - remove any when possible.

## Motivation and Context
- Current typing has wrong type, it occurs type error(following error).
  > JSX element type 'RouteData' does not have any construct or call signatures.
- too many `any` exist in current typing.

I refer https://react-static.js.org/docs/ to write specific types & types of `react-static/node`.

This typing is tested with typescript 3.0.3

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -17,12 +17,23 @@ declare module 'react-static' {
   import { Helmet } from 'react-helmet';
   export class Head extends Helmet {}
 
-  export function withRouteData(comp: any): any;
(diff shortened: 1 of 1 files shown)


Perlmint:
rebased - fix CI failure.

tannerlinsley:
Excellent!

PR #861, opened 2018-11-26
ci: test Node.js 8, 10 and 11

DanielRuf:
## Description

Test against current LTS branches of Node.js. 7 and 9 have already reached their EOL.

## Changes/Tasks

- [ ] Changed code

## Motivation and Context

We should only test on current LTS branches which are used in production as the others do not get any (security) updates anymore.

## Types of changes

## Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: node_js
 node_js:
+  - "node"
   - "10"
-  - "9"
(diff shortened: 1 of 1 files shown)


ScriptedAlchemy:
I’m fine with this. 
We should not focus on EOL legacy

Issue #864, opened 2018-12-03
The problem of using CSS module in react-static


lz-allen:
The Html page packaged in the development environment contains the class name, which will be lost in the production environment.The problem of using CSS module in react-static.

eg:
```
import React from 'react'
import {Link} from 'react-static'
import styles from './panel.less'

const Panel = (props) => {
    return (
        <div className={styles.panel}>
            <h3 className={styles.panelTitle}><span className={styles.text}>{props.title}</span>
                {
                    props.link && <Link className={styles.more} to={props.link}>{props.moreText}</Link>
                }
            </h3>
            {props.children}
        </div>
    )
}
export default Panel
```
## static.config.js
```
webpack: (config, { defaultLoaders, stage }) => {
      let loaders = []

      if (stage === 'dev') {
        loaders = [
          { loader: 'style-loader' },
          {
            loader: 'css-loader',
            options:{
              minimize: stage === 'prod',
              modules: true,
              sourceMap: true,
            }
          },
          { loader: 'less-loader' }]
      } else {
        loaders = [
          {
            loader: 'css-loader',
            options: {
              importLoaders: 1,
              minimize: stage === 'prod',
              modules: true,
              sourceMap: false,
            },
          },
          {
            loader: 'less-loader',
          },
        ]

        // Don't extract css to file during node build process
        if (stage !== 'node') {
          loaders = ExtractTextPlugin.extract({
            fallback: {
              loader: 'style-loader',
              options: {
                sourceMap: false,
                hmr: false,
              },
            },
            use: loaders,
          })
        }
      }

      config.module.rules = [
        {
          oneOf: [
            {
              test: /\.less$/,
              use: loaders,
            },
            {
              test: /\.css$/,
              use: loaders.slice(0, loaders.length - 1).map(item => {
                if(item.loader === 'css-loader'){
                  item ={
                    loader: 'css-loader',
                    options: {
                      importLoaders: 1,
                      minimize: stage !== 'dev',
                      sourceMap: true,
                    },
                  }
                }
                return item
              })
            },
            defaultLoaders.cssLoader,
            defaultLoaders.jsLoader,
            defaultLoaders.fileLoader,
          ],
        },
      ]
      return config
    }
```

nikaspran:
This is a known issue, see https://github.com/nozzle/react-static/issues/601#issuecomment-429574588 for workaround.

lz-allen:
@nikaspran 
[...]

Issue #601, opened 2018-05-07
CSS Modules no longer work


Cinamonas:
As far as I can tell, #574 may have broken the usage of CSS Modules in React-Static. I’ve upgraded to the latest version and noticed that `import styles from 'Foo.scss'` no longer works as expected. If you attempt to reference a class name on the `styles` object, it results in `undefined`.

### Environment

1.  `react-static -v`: 5.8.2 (first non-working release, last working release is 5.8.1)
2.  `node -v`: 8.9.3
3.  `yarn --version or npm -v`: 5.6.0
4.  Operating system: macOS 10.13.4

### Steps to Reproduce the problem

1. Go to [examples/sass](https://github.com/nozzle/react-static/tree/master/examples/sass)
2. `npm i`
3. `npm i react-static@5.8.2` (first non-working release)
4. Open `static.config.js` and add `modules: true` option on `css-loader`
5. Open `App.js` and replace `import './app.scss'` with `import styles from './app.scss'`
6. Add `console.log(styles.content)` below the `app.scss`  import
7. Run `npm stage`
8. Notice that the `console.log` call results in `undefined`

### Expected Behavior

If you did the steps outlined above on 5.8.1, it would work. You’d just need to remove the `if` condition surrounding `ExtractTextPlugin.extract` call in `static.config.js` (which was introduced to conform to changes in #574).

---

Let me know if I’m missing some trivial configuration that would allow me to use CSS Modules. Otherwise, changes in #574 need to be revisited.

/cc @lottamus

JustFly1984:
I already pointed out that it is due to webpack.config.js refactoring, and introducing boolean isNode

lottamus:
It looks like we just need to add `modules: true` to enable CSS modules

https://javascriptplayground.com/css-modules-webpack-react/

Cinamonas:
@lottamus, as pointed out above, I’m adding that. My configuration hasn’t changed, but the CSS Modules no longer work after 5.8.2.

rohmanhm:
Also, I can't get CSS Code Splitting on my app.
The CSS bundles now just compiled into a single CSS file.

sudarshang:
I have a PR which upgrades react-static to webpack 4. Do you want to check if this issue is resolved by using that branch?

https://github.com/nozzle/react-static/pull/605

zackljackson:
https://github.com/faceyspacey/extract-css-chunks-webpack-plugin/tree/webpack-4

[...]

PR #866, opened 2018-12-10
feat: CSS Modules Plugin

ScriptedAlchemy:
## Description

CSS modules plugin added to v6
## Changes/Tasks

- [x] Changed code

## Motivation and Context

People started getting nasty in the git issues. 
It's time that ends

Resolves: #864, #601


tannerlinsley:
The plugins don't currently show in the create command. Just examples. 🙂

tannerlinsley:
Be sure to add it to the docs!

lz-allen:
@tannerlinsley @ScriptedAlchemy 
Thank friends for taking the time to help.

ScriptedAlchemy:
Will wrap docs tomorrow! Don’t want to fumble the ability to pass options through the loader correctly

nikaspran:
@ScriptedAlchemy does this actually work? In my experience, the problem was that RS was not including the `ExtractCssChunks` _plugin_ in the `node` stage, the loaders were actually correct. I'm not familiar with `css-loader/locals` though - does using that mean you don't have to use the ExtractCssChunks plugin any more?

Config change that worked for me, for reference:

  ```diff
  if (stage === 'dev') {
    // Dev
    loaders = [styleLoader, cssLoader]
  } else if (stage === 'node') {
    // Node
  - // Don't extract css to file during node build process
  - loaders = [cssLoader]
  + loaders = [ExtractCssChunks.loader, cssLoader]
  + config.plugins.push(new ExtractCssChunks({
  +   filename: '[name].[chunkHash:8].css',
  +   chunkFilename: '[id].[chunkHash:8].css',
  + }));
  } else {
    // Prod
    loaders = [ExtractCssChunks.loader, cssLoader]
  }
```

Apologies I couldn't do a PR for this myself sooner - I thought the temp. workaround would be enough to fix current issues and I could implement a long term solution once 6.x got a bit more stable. I'll create a PR for 5.x today.

**Comment by ScriptedAlchemy** [comment]
Great question, and it’s one we need to put to the test. I’m basing my solution off of Icarus. Icarus or parts of its architecture currently powering about $800bn (closed source counterparts I own) worth in collectively public companies.  That’s about all I can mention on that side. 

I wrote this on my iPad in the car, so I’ll need to test it tomorrow. I’m banking on if a problem exists it would have cost someone millions by now and they’d have reached out to me 😂 for instance Bloomberg would probs say something if FOUT or busted css was showing up on their modules.  

I do want to test this within RS architecture, but there’s a long list of companies using one of the my architectures, the webpack stuff has been a copy paste for three years or so. Going through the needed reconfigures for WP2,3,4

Not sayin I’m wrong. Just saying it seems to work. 

Once I test it out tomorrow. I’ll give a detailed readme note and comment on the why. Fixing it is one thing, but showing coders new tricks is my jam! 

https://github.com/ScriptedAlchemy/Icarus/blob/master/build/webpack.config.js/loaders.js

**Comment by ScriptedAlchemy** [comment]
From what I know, we don’t actually want to output css files. So running my css extract plugin on node, while harmless is never used. We just want the string that the client is gonna generates thus /locals

It doesn’t actually do anything other than make modules that generate little objects with hashed names. As long as the client has extract css chunks, the class and matches the css files loaded and boom biggidy pob - it should just resume as usual. 

I’m wrong a lot, so I’ll test my PR, and if it’s not working I’ll test ur solution. Whatever fixes it is golden

**Comment by ScriptedAlchemy** [comment]
I think the reason your config work is because extract is actually provisioning files for webpack to consume on the server. We don’t need server css. Just the class and generation. My PR is your work, with an attempt to ask webpack “hey what’s the class name gonna be, gimme that string and skip all that other stuff, your a server and done know how to handle css anyways”

**Comment by ScriptedAlchemy** [comment]
Hey and don’t worry about it. We are all so busy and stretched thin. If we can nail this bug in 5 and 6. It will be an awesome win for morale.

I appreciate your time investigating the bug. I pushed buttons on the keyboard, but you fixed it via TLDR code snippets. 

Respect earned

**Comment by nikaspran** [comment]
I'm not saying your PR won't fix this - just that when I was looking at the issue originally `css-loader/locals` didn't exist yet, but it does sound like exactly what we need.

You're right `ExtractCssChunks` plugin is a bit overkill since we don't actually need the CSS at that stage (removing that unnecessary CSS is the PR that actually broke css modules accidentally), so if this works - perfect!

To help with verifying this - the issue was occurring specifically during `react-static build`, so that's what you'll have to check to make sure it works.

**Comment by ScriptedAlchemy** [comment]
Thanks! I’ll be sure to follow the steps to replicate it! 

And by no means am I defending my solution, like I said I pasted it on my iPad in the car 😂 it seems to be what we need and we will find out soon. All I was saying is I’ve not gotten a complaint so I hope it’s been working in Icarus hehe

Thank you again for your work debugging this. Honestly the credit for this PR is more yours than mine. 

I wish every issue opened contained the diligence of yours in isolating the issue. 

My biggest regret was not seeing your comment sooner! Thanks your your time, ultimately if and when css modules work. 80% is thanks to you 💯💯💯💯💯💯

File: packages/react-static-plugin-css-modules/README.md [hunk 1]
```diff
@@ -0,0 +1,35 @@
...
+## With Options
+
+```javascript
+export default {
+  plugins: [
+    [
+      "react-static-plugin-css-modules",
+      {
+        includePaths: ["..."] // always includes `src/`
+        // other options for the sass-loader
```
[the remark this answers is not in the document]
ScriptedAlchemy: Yeah I need to change the wording and also update the pass through options. It’s 2am here tho, I don’t want to push and screw it up because I’m burnt out. 

I’ll be sure to follow through on your reviews, I know there’s more polish needed for sure

packages/react-static-plugin-css-modules/src/node.api.js
```diff
@@ -0,0 +1,33 @@
...
+        modules: true,
+        importLoaders: 1,
+      }, options),
+    }
+
+    if (stage === 'dev') {
+      // Dev
+      loaders = [styleLoader, cssLoader]
+    } else if (stage === 'node') {
+      cssLoader.loader = 'css-loader/locals'
```
nikaspran: Should this also assign to the `loaders` array? Or maybe something like this in a single action:

```
      loaders = [
        { ...cssLoader, loader: 'css-loader/locals' }
      ]
```

ScriptedAlchemy: Hmm. Technically if it’s node it will always be an object. So we would only need to mutate the loader property, in this instance we are not returning and additional loaders. 

That being said, I likely missed the obvious that in fact we need to re state the loaders variables 🤦‍♂️ 

Yes I believe you are correct. This does indeed need to be updated

tannerlinsley: I think after we get this cleared up, we can merge this.

tannerlinsley:
Also, you'll need to open a new pull request from your fork to `master`...I silently promoted v6 to master today and did a 6.0.0 release. Don't tell yet!

ScriptedAlchemy:
(No disresect)  but YAY! FINALLY 😂😂 long road

PR #872, opened 2018-12-14
Fix issues with windows paths

bkniffler:
See https://github.com/babel/babel/issues/9032

I'm not sure if the ignore path is necessary. But I know it will cause issues in its current form. I'm not able to start any react-static example without this change.

## Types of changes

### DIFF
--- a/packages/react-static/package.json
+++ b/packages/react-static/package.json
@@ -66,6 +66,7 @@
     "inquirer": "^6.1.0",
     "inquirer-autocomplete-prompt": "^1.0.1",
     "intersection-observer": "^0.5.1",
+    "lodash": "^4.17.11",
     "match-sorter": "^2.2.3",
     "minimist": "^1.2.0",
     "mutation-observer": "^1.0.3",
--- a/packages/react-static/package.json
+++ b/packages/react-static/package.json
@@ -66,7 +66,6 @@
     "inquirer": "^6.1.0",
     "inquirer-autocomplete-prompt": "^1.0.1",
     "intersection-observer": "^0.5.1",
-    "lodash": "^4.17.11",
     "match-sorter": "^2.2.3",
     "minimist": "^1.2.0",
     "mutation-observer": "^1.0.3",
--- a/packages/react-static/src/utils/binHelper.js
+++ b/packages/react-static/src/utils/binHelper.js
@@ -1,9 +1,21 @@
+const path = require('path');
+const escapeRegExp = require('lodash/escapeRegExp');
 let ignorePath
 
 // Allow as much stack tracing as possible
 Error.stackTraceLimit = Infinity
 
-require('@babel/register')({})
+require('@babel/register')({
+  ignore: [
+    function babelIgnore(filename) {
+      // true if should ignore
+      return (
+        new RegExp(escapeRegExp(`${path.sep}node_modules${path.sep}`)).test(filename) ||
+        (ignorePath && ignorePath.test(filename))
+      )
+    },
+  ],
+});
 
 const updateNotifier = require('update-notifier')
 const PrettyError = require('pretty-error')
@@ -53,6 +65,6 @@ process.on('unhandledRejection', r => {
 
 module.exports = {
   setIgnorePath(path) {
-    ignorePath = path ? new RegExp(path) : undefined
+    ignorePath = path ? new RegExp(escapeRegExp(path)) : undefined
   },
 }
--- a/packages/react-static/src/utils/binHelper.js
+++ b/packages/react-static/src/utils/binHelper.js
@@ -3,17 +3,7 @@ let ignorePath
 // Allow as much stack tracing as possible
 Error.stackTraceLimit = Infinity
 
-require('@babel/register')({
-  ignore: [
-    function babelIgnore(filename) {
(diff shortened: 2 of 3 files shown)


tannerlinsley:
My guess is the error is being caused by something else, but I'd like to know more about this error as well.

bkniffler:
I did some more research and found that this is a problem with unix vs windows path (using windows currently). The original ignore took only unix into consideration.

Here is the original ignore option of babel:
https://github.com/babel/babel/blob/47da5cf75a4f67d199422e7f2a2635a71317d478/packages/babel-register/src/node.js#L145

The windows path style causes problems in other places too. I'm still getting `Error: Could not find a route for the "index" page of your site! This is required. Please create a page or specify a route and template for this page.` when running the examples which I've been talking about here: https://github.com/nozzle/react-static/issues/721

bkniffler:
A bit off-topic, I also notices that `getRoutesFromPages` in installed `react-static` lib folder contains 

```
(function () {
  var reactHotLoader = require('react-hot-loader').default;

  var leaveModule = require('react-hot-loader').leaveModule;

  if (!reactHotLoader) {
    return;
  }

  reactHotLoader.register(watcher, "watcher", "/Users/tannerlinsley/GitHub/react-static/packages/react-static/src/static/getRoutes/getRoutesFromPages.js");
  reactHotLoader.register(routesCache, "routesCache", "/Users/tannerlinsley/GitHub/react-static/packages/react-static/src/static/getRoutes/getRoutesFromPages.js");
  leaveModule(module);
})();
```

bkniffler:
Okay, I can run example projects on windows with the changes applied here. I don't know if there is still problems due to windows path vs unix path, it is difficult to find them as long as they don't cause exceptions.

You should definitely use nodejs `path.resolve` / `path.sep` etc more, hard-coding '/' in `str.replace` or `str.split` is super error prone.

bkniffler:
Lodash is already part of many dependencies of react-static.
```
info Reasons this module exists
   - "@babel#types" depends on it
   - Hoisted from "@babel#types#lodash"
   - Hoisted from "@babel#traverse#lodash"
   - Hoisted from "@babel#generator#lodash"
   - Hoisted from "react-static#webpack-bundle-analyzer#lodash"
   - Hoisted from "react-static#inquirer#lodash"
   - Hoisted from "react-static#@babel#cli#lodash"
   - Hoisted from "react-static#@babel#core#lodash"
   - Hoisted from "react-static#@babel#register#lodash"
   - Hoisted from "react-static#css-loader#lodash"
   - Hoisted from "react-static#html-webpack-plugin#lodash"
   - Hoisted from "eslint-config-react-tools#eslint#lodash"
   - Hoisted from "eslint-config-react-tools#eslint-plugin-import#lodash"
   - Hoisted from "react-static#extract-css-chunks-webpack-plugin#lodash"
   - Hoisted from "react-static#react-dev-utils#inquirer#lodash"
   - Hoisted from "react-static#thread-loader#async#lodash"
   - Hoisted from "react-static#@babel#preset-env#@babel#plugin-transform-block-scoping#lodash"
   - Hoisted from "eslint-config-react-tools#eslint#table#lodash"
   - Hoisted from "react-static#webpack-dev-server#http-proxy-middleware#lodash"
   - Hoisted from "react-static#optimize-css-assets-webpack-plugin#last-call-webpack-plugin#lodash"
   - Hoisted from "eslint-config-react-tools#babel-eslint#@babel#types#lodash"
   - Hoisted from "eslint-config-react-tools#babel-eslint#@babel#traverse#lodash"
   - Hoisted from "eslint-config-react-tools#eslint#inquirer#lodash"
   - Hoisted from "react-static#@babel#preset-env#@babel#plugin-transform-classes#@babel#helper-define-map#lodash"
   - Hoisted from "react-static#@babel#preset-env#@babel#plugin-transform-modules-umd#@babel#helper-module-transforms#lodash"
   - Hoisted from "react-static#@babel#preset-env#@babel#plugin-proposal-unicode-property-regex#@babel#helper-regex#lodash"
   - Hoisted from "eslint-config-react-tools#babel-eslint#@babel#traverse#@babel#generator#lodash"
   - Hoisted from "eslint-config-react-tools#babel-eslint#@babel#traverse#@babel#helper-function-name#@babel#template#lodash"
```

packages/react-static/package.json
```diff
@@ -66,6 +66,7 @@
     "inquirer": "^6.1.0",
     "inquirer-autocomplete-prompt": "^1.0.1",
     "intersection-observer": "^0.5.1",
+    "lodash": "^4.17.11",
     "match-sorter": "^2.2.3",
```
tannerlinsley: This can be removed now.

packages/react-static/src/utils/binHelper.js
```diff
@@ -1,3 +1,6 @@
+const path = require('path');
+const escapeRegExp = require('lodash/escapeRegExp');
```
tannerlinsley: You can just import this from the general Utils now:
```javascript
const { escapeRegExp } = require('./')
```

bkniffler:
Done!

ScriptedAlchemy:
Thanks for the hard work and helping us understand!

PR #885, opened 2018-12-17
less aggressive prefetching (check if link should be prefetched)

digitalkaoz:
## Description

we do some checks now if the link should be prefetched et all.

## Changes/Tasks

- [x] Changed code

## Motivation and Context

a lot of avoidable `routeInfo.json` requests are dropped which results in `404` responses (to e.g. external urls or `javascript:` links)

## Types of changes

## Checklist:

### DIFF
--- a/packages/react-static/src/browser/utils/index.js
+++ b/packages/react-static/src/browser/utils/index.js
@@ -198,3 +198,39 @@ export function getBasePath() {
     ? ''
     : process.env.REACT_STATIC_BASE_PATH
 }
+
+export function isPrefetchableRoute(path) {
+  // when rendering static pages we dont need this et all
+  if (isSSR()) {
+    return false;
+  }
+
+  // script links
+  if (path.indexOf('javascript:') === 0) {
+    return false;
+  }
+
+  const self = document.location;
+  let link;
+
+  try {
+    link = new URL(path);
+  } catch (e) {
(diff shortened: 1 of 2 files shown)


packages/react-static/src/browser/utils/index.js
```diff
@@ -198,3 +198,23 @@ export function getBasePath() {
...
+  // script links
+  if (path.indexOf('javascript:') === 0) {
+    return false;
+  }
+
+  try {
+    new URL(path);
+  } catch (e) {
+    // if a path is not parsable by URL its a local relative path
+    return true;
```
tannerlinsley: Do you know if this breaks when using a `config.siteRoot`? When using that, all links are prefixed to be absolute to the site and would technically be a full URL.

tannerlinsley: `getRoutePath` should actually strip that though... So my guess is everything is fine with this, but I just wanted you to confirm ;)

digitalkaoz:
@tannerlinsley you were right. catched this case too. @paavohuhtala your case should be fixed too. test at your own will if all cases are handled correctly. i think it should be fine now...

digitalkaoz:
@tannerlinsley thanks for merging and fixing the CS :)

tannerlinsley:
You bet :) Thanks for the great PR!

PR #899, opened 2018-12-19
Css modules

ScriptedAlchemy:
## Changes/Tasks

- [ ] Changed code

## Types of changes

## Checklist:


tannerlinsley:
Awesome. FYI you don't need to increment the version. Lerna does that for us on release. We just need to start marking our commits and PRs with patch, minor or major. Thoughts?

ScriptedAlchemy:
I installed the latest version but implemented the previous versions method

ScottSmith95:
This is working as expected for me.

ScottSmith95:
Do you think this is compatible with adding PostCSS loader after CSS Modules? It seems that just adding them after cssLoader and adding `importLoaders: 2` isn't working alone.

ScriptedAlchemy:
@ScottSmith95 it should, just like any other webpack loader, I’d imagine it should work with postcss. 

Can you throw some code up to me and I’ll check ur config

ScriptedAlchemy:
Going to merge this, we can open more pill requests as needed but as it seems to solve the base issue of css modules. Let’s get it in master

ScottSmith95:
@ScriptedAlchemy If you wouldn't mind, I have tried creating a plugin for Sass and CSS Modules and I"m finding it that many styles don't appear on the build stage, although they do in development. Any idea what in this config could be causing the issue? 
```javascript
import ExtractCssChunks from "extract-css-chunks-webpack-plugin";

export default ({ cssLoaderOptions = {}, sassIncludePaths = [] }) => ({
  webpack: (config, { stage }) => {
    let loaders = [];

    const styleLoader = {
      loader: "style-loader",
      options: { sourceMap: true }
    };
    const cssLoader = {
      loader: "css-loader",
      options: {
        importLoaders: 2,
        modules: true,
        camelCase: true,
        localIdentName:
          stage === "dev"
            ? "[name]__[local]--[hash:base64:5]"
            : "[hash:base64:5]",
        sourceMap: true,
        ...cssLoaderOptions
      }
    };
    const postcssLoader = {
      loader: "postcss-loader",
      options: {
        config: {
          // File must be named in one of these formats:
          // https://github.com/michael-ciniawsky/postcss-load-config#usage
          path: "src/configuration/"
        }
      }
    };
    const sassLoader = {
      loader: "sass-loader",
      options: {
        sassIncludePaths: ["src/", ...sassIncludePaths],
        sourceMap: true
      }
    };

    if (stage === "dev") {
      // Dev
      loaders = [styleLoader, cssLoader, postcssLoader, sassLoader];
    } else if (stage === "node") {
      // Node
      // Don't extract css to file during node build process
      loaders = [
        {
          ...cssLoader,
          loader: "css-loader",
          options: { exportOnlyLocals: true }
        },
        postcssLoader,
        sassLoader
      ];
    } else {
      // Prod
      config.entry = ["@babel/polyfill", ...config.entry];
      loaders = [ExtractCssChunks.loader, cssLoader, postcssLoader, sassLoader];
    }

    config.module.rules[0].oneOf.unshift({
      test: /\.s(a|c)ss$/,
      use: loaders
    });

    return config;
  }
});

```

PR #904, opened 2018-12-19
Fix name of `jsLoaderExt` in the plugins API

Vinnl:
## Description

The key is actually `jsLoaderExt`, rather than `jsLoaderExternal`, as can be seen here:
https://github.com/nozzle/react-static/blob/master/packages/react-static/src/static/webpack/rules/index.js#L8

Alternatively, the key could be changed there, but that would be a breaking change. Your call :)

## Changes/Tasks

- [ ] Fixed name of `jsLoaderExt` in documentation

## Motivation and Context
The documentation was incorrect w.r.t. the actual code.

## Types of changes

## Checklist:

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -123,7 +123,7 @@ webpack: []Function(
     stage,
     defaultLoaders: {
       jsLoader,
-      jsLoaderExternal,
+      jsLoaderExt,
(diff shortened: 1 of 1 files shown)


tannerlinsley:
Good catch!

PR #908, opened 2018-12-20
Add react-static-plugin-typescript to plugin list

Vinnl:
## Description

I created react-static-plugin-typescript; this PR adds it to the list of plugins.

Note: on Twitter I mentioned wanting to give this a more thorough run-through on my own project, but since I'm [stuck on upgrading that](https://spectrum.chat/react-static/general/how-to-use-reach-router-with-react-static-v6~c0b4658e-bd3c-4d15-93da-8391a491b9dc) I figured I'd submit this anyway, since it otherwise appears to work fine.

## Changes/Tasks

- [x] Updated plugin list in README

## Motivation and Context
It tells people who want to use TypeScript which plugin they can use.

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,8 @@ Once you've installed and test driven, you may want to:
   - [react-static-plugin-sass](/packages/react-static-plugin-sass) - Adds SSR and general support for SASS
 - React Alternatives
   - [react-static-plugin-preact](/packages/react-static-plugin-preact) - Adds preact support
+- Type checking
+  - [react-static-plugin-typescript](https://www.npmjs.com/package/react-static-plugin-typescript) - Allows you to write your components in TypeScript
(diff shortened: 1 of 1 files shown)


digitalkaoz:
@Vinnl how far are you using typescript with react-static?
Im failing use typescript in every aspect:

`static.config.ts` wont work (babel is not already in place here)
`src/index.tsx` wont work (file resolutions is weak here)

do you have an example repo where you use typescript completely?

Vinnl:
@digitalkaoz I'm only using it for my own app code, so the React Static config is still `static.config.ts`. I think #978 should make it possible to use a `.ts` extension for your React Static config _if_ you're running it using `ts-node`, but I haven't tried that myself.

An example can be seen if you initialise a new React Static application using the TypeScript template. That doesn't enable TypeScript for `static.config.js`, but does run successfully for all the rest.

PR #914, opened 2018-12-23
let webpack resolve with absolute paths in case of an absolute output…

digitalkaoz:
there was a bug (recently introduced?, it worked with an older react-static beta) when building to an absolute `DIST` path. in case of an absolute `DIST` path we simply tell webpack to resolve all modules with absolute paths.

@tannerlinsley this is relativly urgent :)

### DIFF
--- a/packages/react-static/src/static/buildHTML.js
+++ b/packages/react-static/src/static/buildHTML.js
@@ -15,7 +15,7 @@ export default (async function buildHTML({
   time(chalk.green('=> [\u2713] HTML Exported'))
 
   // in case of an absolute path for DIST we must tell node to load the modules from our project root
-  if (config.paths.DIST.indexOf(config.paths.ROOT) !== 0) {
+  if (!config.paths.DIST.startsWith(config.paths.ROOT)) {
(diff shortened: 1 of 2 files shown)


ScriptedAlchemy:
Ive not worked on the modules section of RS. However, given my Webpack knowledge these resolvers make sense. @tannerlinsley depending on our test coverage, it would be worth releasing this or generating a `next` tag on NPM to allow the community to test our patches before we roll them. 

This is a practice i have done for React Static many times with Universal and Extract CSS Chunks. Module resolution is always something I try to release gently as a beta. Especially on large projects 

Your call, perhaps I’m being over cautious. You know the architecture of the Lerna repo better than myself

digitalkaoz:
@ScriptedAlchemy as most users dont build to some external paths (its only needed in readonly filesystems, e.g. Lambda) and we keep the current resolution for relative dist builds, i see No Problem. For the ones (me) building to an external path, its broken anyway...

tannerlinsley:
I'm okay releasing if it's broken. I'll get on this in the morning.

ScriptedAlchemy:
thanks for clarifying. with this context - you’ve got my blessing.

PR #925, opened 2018-12-28
Make sure plugin options are passed on to plugins

Vinnl:
## Description
Options can be provided to plugins by passing an array instead of
the plugin name, with as first element the plugin name, and an
options object as the second argument. However, this array was
destructured as follows:

      originalLocation = originalLocation[0]
      options = originalLocation[1] || {}

Because the first line overwrote `originalLocation` with the name
of the plugin, the second line would actually assign the second
character of that name to `options`, rather than the second element
of the array. Switching them around fixes that.

## Changes/Tasks

- [x] Changed code
- [x] Added tests (see notes at the bottom)

## Motivation and Context
Plugins could not actually receive options. This PR fixes that.

## Types of changes

## Checklist:

Note that I am not too satisfied with the way I structured the tests, especially the part where I modified the actual code to set a plug-in's name to `mock-plugin` when in the test environment (as opposed to undefined as used to be the case). I then mocked `extra-fs` to return `true` when checking whether the `mock-plugin` plugin exists, so that I could provide a mock for that plugin.

Unfortunately, I couldn't think of a better way to test this. If anyone knows of a better way I'm all ears, but otherwise I guess we'll have to make do with this.

### DIFF
--- a/packages/react-static/package.json
+++ b/packages/react-static/package.json
@@ -130,7 +130,8 @@
     "testRegex": "(/__tests__/.*\\.(test))\\.jsx?$",
     "moduleNameMapper": {
       "static.config.js": "<rootDir>/src/static/__mocks__/static.config.mock.js",
-      "./path/to/static.config.js": "<rootDir>/src/static/__mocks__/static.config.mock.js"
+      "./path/to/static.config.js": "<rootDir>/src/static/__mocks__/static.config.mock.js",
+      "./path/to/configWithPluginWithOptions.mock.js": "<rootDir>/src/static/__mocks__/configWithPluginWithOptions.mock.js"
     },
     "setupFiles": [
       "<rootDir>/setupTests.js"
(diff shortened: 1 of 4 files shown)


tannerlinsley:
This is perfectly sufficient for now. You rock!

Vinnl:
Thanks Tanner! One more question: at what point do you cut new releases? This and #912 are the only roadblocks to [allowing for actual type checking](https://gitlab.com/Vinnl/react-static-plugin-typescript/merge_requests/1) in the TypeScript plugin :)

(No rush! Just trying to get a better picture of my own timelines.)

tannerlinsley:
I'll do a release by the end of the day. Maybe sooner.

Vinnl:
Ah, so very soon. Again, no rush - it's unlikely that I'll be able to get to it this weekend anyway :)

PR #935, opened 2019-01-02
fix css-loader 2.0 support

ScottSmith95:
## Description
CSS Loader 2.0 disables CSS Modules by default, so in the node build config, css module support is disabled and no mappings are exported. The `options` object needs to remain present in the node config, albeit with the addition of `exportOnlyLocals: true`.

## Changes/Tasks
- Added `...options` to css-loader options during node build stage.

## Motivation and Context
With css-loader v2 builds break without this change.
**No open issue for this**

## Types of changes

## Checklist:


ScottSmith95:
@ScriptedAlchemy this resolves the issue referenced here: https://github.com/nozzle/react-static/pull/899#issuecomment-448722819

PR #941, opened 2019-01-04
Update animated-routes example to animate properly.

nicolas-cusan:
## Description
The example does not work with the latest version of react-sprint. See react-spring `Transition` usage http://react-spring.surge.sh/transition#/red The PR fixes the issue and the animation works properly. 

There is still an issue with `withRouteData` & `withSiteData` when navigating as the route that is on its way out looses its data due to a re-render. I was not able to fix that yet.

## Changes/Tasks

Fixed animated-routes example code to work with latest react-spring version

## Motivation and Context

The animated-routes example did not work with the latest version of react-sprint.

## Types of changes

## Checklist:

### DIFF
--- a/docs/guides/animated-routes.md
+++ b/docs/guides/animated-routes.md
@@ -9,7 +9,7 @@ Animated Routes can be achieved so many different ways. In this example, we'll s
 import React from 'react'
 import { Root, Routes } from 'react-static'
 import { Link } from '@reach/router'
-import { Transition } from 'react-spring'
+import { Transition, animated } from 'react-spring'
 
 const App = () => (
   <Root>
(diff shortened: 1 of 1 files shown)


marqpdx:
hello nicolas-cusan

thx for this patch. 

you also wrote: _"There is still an issue with `withRouteData` & `withSiteData` when navigating as the route that is on its way out looses its data due to a re-render. I was not able to fix that yet."_

i believe i am having the same issue; when i navigate away from the blog page, i get an error `TypeError: Cannot read property 'map' of undefined` and it's pointing to `{posts.map(post => (...` and we get the error, i surmise, b/c it's on its way out of the blog page and loses its data. 

have you made any progress on that?

i was thinking simply to check to make sure posts is not undefined, but not exactly sure the syntax.

thx,
m

nicolas-cusan:
Hello @marqpdx, I did not fix the issue and ended up using another tool for the job.

The problem with just a check for `undefined` ist that, as the _leaving_ page is re-rendered, you will have no data to show while it is on its way out. The posts (or whatever data it had via `withRouteData`) would just disappear and then the out-animation would start.

I tried to somehow keep track of the old data but it did not work out. If you find a solution please let me know, I really like the this tool and having the page transitions work would be great.

Thx, Nico

PR #959, opened 2019-01-10
Update react-dom to 16.7 alpha

NavyAdmiral:
I see react was updated to ^16.7.0-alpha.2 in a commit that mentioned hooks. Shouldn't react-dom have been updated correspondingly too?


tannerlinsley:
Yes.

Astrantia:
@tannerlinsley Can we get a release for this?

PR #963, opened 2019-01-11
add typescript template

sw-yx:
## Description

add typescript template. mostly a direct port of the basic template to typescript, but with aliases set up properly.

## Motivation and Context

make it easier to use RS with TS

## Types of changes

## Checklist:


tannerlinsley:
I think the plugin takes care of some more things you have manually defined. Might want to check: https://gitlab.com/Vinnl/react-static-plugin-typescript/blob/master/src/node.api.ts

sw-yx:
ok i can remove `extensions: ['.js', '.jsx', '.ts', '.tsx']` 

from what i can tell that is the only line i can take out tho... unless i simplify the `webpack.config.js` insertion process

tannerlinsley:
That's your call. Anything to make it more reusable.

sw-yx:
alright simplified a buncha stuff

tannerlinsley:
What else could we do to avoid the API file?

sw-yx:
are they a bad thing? theyre the canonical place to modify webpack config right?

tannerlinsley:
Yeah but if it's typescript related it could possibly be passed as a plugin
option.

tannerlinsley:
If you're reasonably convinced this is how far we can push things into the plugin, then let me know and we'll get this merged.

sw-yx:
sorry been busy with the netlify allhands past few days. gonna try a few things and see - my lack of complete knowledge of the RS api (and imperfect knowledge of webpack, its always been a weak point of mine) is probably hindering me here from a totally elegant solution but we can definitely improve this over time.

already used it for my podcasting site btw

sw-yx:
ok so i figured I was only using the node.api for path aliases which are frankly optional so i could just take them out. if it were to be a plugin option it would have to be added upstream to `react-static-plugin-typescript` by @Vinnl.

anyway - removing them made it a lot simpler, and brings us closer to just doing what the basic template does but in typescript!

packages/react-static/templates/typescript/package.json
```diff
@@ -0,0 +1,38 @@
...
+  },
+  "devDependencies": {
+    "@types/node": "^10.12.18",
+    "@types/reach__router": "^1.2.2",
+    "@types/react": "^16.7.18",
+    "@types/react-dom": "^16.0.11",
+    "@types/react-hot-loader": "^4.1.0",
+    "@types/webpack-env": "^1.13.6",
+    "convert-tsconfig-paths-to-webpack-aliases": "^0.9.2",
+    "ts-loader": "^5.3.3",
```
Vinnl: If you add the plugin, you shouldn't be required to install TypeScript and a loader yourself. (And in any case, the plugin doesn't use ts-loader :) )

packages/react-static/templates/typescript/tsconfig.json
```diff
@@ -0,0 +1,25 @@
...
+    "noUnusedParameters": true,
+    "removeComments": false,
+    "preserveConstEnums": true,
+    "sourceMap": true,
+    "skipLibCheck": true,
+    "baseUrl": ".",
+    "paths": {
+      "@components/*": ["src/components/*"]
+    },
+    "typeRoots": ["./node_modules/@types"],
```
Vinnl: I don't think `typeRoots` is necessary? For reference, see the tsconfig at https://www.npmjs.com/package/react-static-plugin-typescript#usage

sw-yx:
ohhh nice! ok that can work. and thanks for the tips, will update

Vinnl:
> I could work on [alias support] this weekend

It was more involved than I estimated it to be, but on the plus side, I did most of the work tonight. I still have to clean it up for actual release, but it'd be great if you could test it using the template before that. You can install it using:

    npm install --save-dev react-static-plugin-typescript@1-support-aliases

sw-yx:
wowww. i'm so glad i tagged you lol. trying now.

sw-yx:
@Vinnl it works! waiting for whenever you are ready to update your plugin, then i'll push/merge this template with the final version

Vinnl:
Great! Coincidentally I just finished adding tests and such. It's building now, when that's finished I'll release 3.1.0 with alias support. I'll ping you when that's done :)

(Don't forget to update the README btw! :) )

Vinnl:
@sw-yx It's been released! V3.1.0.

tannerlinsley:
This is great you guys. Seriously, so good.

tannerlinsley:
The only thing left is to add it to this page: https://github.com/nozzle/react-static/tree/master/packages/react-static/templates

sw-yx:
ok, updated and added a mention in that README

tannerlinsley:
Merged! Thanks guys. This is just awesome.

tannerlinsley:
It should now be available in the latest release 🎉

Vinnl:
Great, thanks! Unfortunately, I did still run into an error when trying it out:

```
$ react-static start
  Error: /tmp/my-static-site-ts/static.config.js:1
  (function (exports, require, module, __filename, __dirname) { import axios fro  m 'axios';
                                                                       ^^^^^
  SyntaxError: Unexpected identifier
```

This is in `static.config.js`, so not a TypeScript file. However, it does not occur with the basic template. Any of you have any idea what that could be?

sw-yx:
@Vinnl i can replicate this. i've no idea why this is happening since axios is in package.json. very very puzzling.

shamrin:
I've tried typescript template, but it failed right away: #989.

ngocdaothanh:
TypeScript template does not support image import:
https://github.com/react-static/react-static/issues/1589

I guess it's missing Webpack config?

SleeplessByte:
Thanks for opening a new issue. It's not good practice to comment on a PR merged over 2 years ago.

PR #973, opened 2019-01-16
Add prefetch disabling for specific Links

krzysztof-pyrzyk-ah:
## Description
I've added one more check before prefetch is being done so that it can be disabled easily.
This feature was working fine in React-Static v5, but in v6 there's a new router it stopped working.

By adding `prefetch="false"` to a `<Link>` component we can prevent React-Static from prefetching this route.

Some documentation has to be added, but I wasn't sure where to put it.
Waiting for your suggestions.

## Motivation and Context
Without this change you cannot disable prefetching specific links.
It fixes an issue I've opened [here](https://github.com/nozzle/react-static/issues/970).

## Types of changes

## Checklist:


tannerlinsley:
Genius :)

PR #978, opened 2019-01-19
Support non-js extensions (.jsx, .mjs, .ts, etc.)

JamesMessinger:
## Description
This PR enhances the `getConfig.js` file to support non-js file extensions for all of the React Static config files:

- `static.config.js`
- `node.api.js`
- `browser.api.js`
- `src/index.js`

All of these files are now loaded via a function called `resolveModule()`, which first attempts to load the module via Node's built-in `require.resolve()`.  This will automatically support any module extension that's built into Node (i.e. `.js`, `.mjs`, `.node`, etc).  It will also support any extensions that have been registered via Node require hooks, such as `.jsx`, `.ts`, etc.

If `require.resolve()` is unable to find the module, then the `resolveModule()` function falls-back to `fs-extra` via `fs.pathExistsSync()`.  Even this fallback approach is enhanced though.  It now supports any of the extensions in `config.extensions`, rather than just `.js`.

## Changes/Tasks
- [x] Enhanced `getConfig.js` as described above
- [x] Added a new test to `getConfig.test.js` that loads a `static.config.jsx` file using the functionality described above
- [x] Added a mock `static.config.jsx` file that includes a React component (the `Document` option)

## Motivation and Context
This change allows people to use whatever file extensions they prefer (i.e. `.jsx` instead of `.js`).  This is actually required by some IDEs and text editors to get correct code highlighting and syntax support.  Many text editors only support JSX syntax in `.jsx` files.

It also adds support for TypeScript via [ts-node](https://www.npmjs.com/package/ts-node).  TS-Node is just a pass-through to Node.js, but it registers a require hook that adds support for TypeScript (`.ts` and `.tsx`) files.  Users can now seamlessly use TypeScript with React Static by using files like `static.config.ts` and `src/index.tsx`.

## Types of changes

## Checklist:

### DIFF
--- a/packages/react-static/package.json
+++ b/packages/react-static/package.json
@@ -129,6 +129,7 @@
     "testRegex": "(/__tests__/.*\\.(test))\\.jsx?$",
     "moduleNameMapper": {
       "static.config.js": "<rootDir>/src/static/__mocks__/static.config.mock.js",
+      "./path/to/static.config": "<rootDir>/src/static/__mocks__/static.config.mock.jsx",
       "./path/to/static.config.js": "<rootDir>/src/static/__mocks__/static.config.mock.js",
       "./path/to/configWithPluginWithOptions.mock.js": "<rootDir>/src/static/__mocks__/configWithPluginWithOptions.mock.js"
     },
--- a/packages/react-static/src/static/__mocks__/static.config.mock.jsx
+++ b/packages/react-static/src/static/__mocks__/static.config.mock.jsx
@@ -1,0 +1,13 @@
+import * as React from "react";
+
+export default {
+  Document: ({ Html, Head, Body, children }) => (
+    <Html lang="en-US">
+      <Head>
+        <meta charSet="UTF-8" />
+        <meta name="viewport" content="width=device-width, initial-scale=1" />
+      </Head>
+      <Body>{children}</Body>
+    </Html>
+  ),
+}
--- a/packages/react-static/src/static/__tests__/getConfig.test.js
+++ b/packages/react-static/src/static/__tests__/getConfig.test.js
@@ -124,6 +124,14 @@ describe('getConfig', () => {
       })
     })
 
+    it('should find the configuration file using any supported extension', async () => {
+      // mapped by the moduleNameMapper in package.json -> src/static/__mocks__/static.config.jsx
+      const configuration = await getConfig('./path/to/static.config')
+
+      testConfiguration(configuration, defaultConfigProduction)
+      expect(configuration.Document).toBeInstanceOf(Function) // React component
+    })
+
     it('should pass on plugin options to those plugins', async () => {
       mockPlugin.mockReset()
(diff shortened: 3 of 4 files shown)


tannerlinsley:
Any thing you need from me for passing tests? This is sooooooo good!!!!

JamesMessinger:
It's weird, the tests pass fine for me on my local machine.  I even tried on both Windows and Ubuntu.  Both worked fine.   Not sure what's different in Travis-CI  😕

Can you pull the PR branch and see if the tests pass on your local machine?

tannerlinsley:
I honestly don't know. I am very inexperienced with testing. 😔😔😔

sw-yx:
james this is so good. very excited for this!

tannerlinsley:
I really want to merge this. Anyone have any ideas on the testing issue?

digitalkaoz:
@JamesMessinger its still trying to load `src/index.js` here. i only have `src/index.tsx` all other files are loadable with their Typescript counterparts.

can you recheck that?

JamesMessinger:
@digitalkaoz - The code first tries to resolve the file using Node's built-in resolution algorithm (`require.resolve()`).  If that fails, then it falls back to the `config.extensions` setting, which defaults to `.js` and `.jsx`.   Here's the relevant code:

https://github.com/nozzle/react-static/blob/fa89ffdead2d2da871ce98e6715518e628faea52/packages/react-static/src/static/getConfig.js#L307-L317

Neither React Static nor Node supports TypeScript by default.  So you'll need to use [TS Node](https://github.com/TypeStrong/ts-node) instead.  TS Node enhances Node's built-in `require.resolve()` algorithm to support TypeScript extensions (`.ts` and `.tsx`).  So the code above ☝️ should find your `index.tsx` file in the first pass.

If, for some reason, that doesn't work, then you can force it to find your `index.tsx` file in the _second_ pass by setting the `config.extensions` property to `[".tsx", ".ts", ".jsx", ".js"]`.

digitalkaoz:
@JamesMessinger  yeah, i guessed that. sadly setting `DEFAULT_EXTENSIONS` to `[".tsx", ".ts", ".jsx", ".js"]` 
fails way earlier now:

```
Error: .../static.config.ts:1
  (function (exports, require, module, __filename, __dirname) { import { readFileSync } from "fs";
                                                                       ^
  SyntaxError: Unexpected token {
```

if i set it in the `static.config.js` i still get `Module not found: Can't resolve '.../src/index.js' in '.../node_modules/react-static/node_modules'` 

i think ts-node is not required at all since im using the typescript-babel-plugin with the `react-static-plugin-typescript`.

there is definitly some bug in here using typescript. I simply cant point my finger on it yet...

digitalkaoz:
@JamesMessinger ok here is the bug:

`console.log(path, config && config.extensions);` in `resolveModule` of `getConfig.js`
while resolving `index` and `static.config` we dont pass the parsed config (how ;) at least for `static.config`) but for `index` and might already work

```
.../static.config undefined
.../index undefined
.../node.api [ '.js', '.jsx' ]
.../browser.api [ '.js', '.jsx' ]
```

PR #979, opened 2019-01-19
Added TypeScript definitions for `static.config.js` files

JamesMessinger:
## Description
This PR just adds TypeScript definitions for the `static.config.js` file.

## Changes/Tasks
- [x] TypeScript definitions

## Motivation and Context
For users who are writing their `static.config` file in TypeScript, these definitions add type checking, code completion, etc.

For users who are writing their `static.config` file in plain JavaScript, but using a TypeScript-aware code editor (such as VSCode), these definitions add code-completion.  If they're using [`// @ts-check`](https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html), then they'll also get type checking.

## Types of changes

## Checklist:

### DIFF
--- a/packages/react-static/src/index.d.ts
+++ b/packages/react-static/src/index.d.ts
@@ -8,6 +8,10 @@
 
 declare module 'react-static' {
   import * as React from 'react'
+  import { ComponentType, ReactNodeArray } from "react";
+  import { Configuration as WebpackDevServerConfig } from "webpack-dev-server";
+
+  type AnyReactComponent = ComponentType<Record<string, any>>;
 
   // Passing on helmet typings as "Head"
   import { Helmet } from 'react-helmet'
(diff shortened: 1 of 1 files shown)


tannerlinsley:
You've done it again. 🎉🎉🎉

Vinnl:
Awesome @JamesMessinger! One potentially possible improvement is to allow typing plugin configurations. For example, my plugins already have type definitions for the available options, so perhaps it would be possible to allow for extending the plugin config from these type definitions. I'm not _that_ familiar with writing type definitions yet, but I believe that that is possible?

I'd imagine something like this:

```typescript
  export type ConfiguredPlugin = [string, object];

  export interface ReactStaticConfig {
    // ...
    plugins?: Array<string | ConfiguredPlugin>;
```

and then a plugin would be able to add (guessing at the syntax here):

```typescript
  declare module 'react-static' {
    ConfiguredPlugin = ['some-plugin', SomePluginOptions];
  }
```
(the idea being that it extends the existing definition of `ConfiguredPlugin` - not sure if that's how this actually works.)

What do you think?

JamesMessinger:
@Vinnl - That's a tricky one.  The idea you suggested will work (it's called [module augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation)), **but** if multiple different plugins use this same technique, then they'll all be fighting over the definition of the `ConfiguredPlugin` type.

PR #983, opened 2019-01-22
Add switch to the Routes component to disable force update on location change. 

georgehenderson:
This PR adds a switch to the Routes component to disable force update on location change. This is useful for cases where you bring your own router and don't need the Routes component to force an update with url changes.

## Description
Adds a new default prop `disableUpdateOnLocationChange: false`

## Changes/Tasks
- [x] New default prop `disableUpdateOnLocationChange: false`

## Motivation and Context
The Routes component causes the next page to mount twice when inside a Reach Router. This can lead to bugs when doing working in `componentDidMount` like logging analytics. 
The router setup for reference
```javascript
<Router>
  <Routes default />
</Router>
```

## Types of changes

## Checklist:

### DIFF
--- a/docs/components.md
+++ b/docs/components.md
@@ -63,6 +63,12 @@ export default () => (
 
 The routes that will be rendered are the **routes** returned by the `getRoutes` function of this config.
 
+#### Props
+
+| Prop                   | type      | Default | description                                                       |
+| ---------------------- | --------- | ------- | ----------------------------------------------------------------- |
+| `disableUpdateOnLocationChange`      | `bool`    | false   | Toggles location changes from updating the component tree                                    |
+
 ##### Custom `Routes` Rendering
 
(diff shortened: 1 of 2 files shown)


georgehenderson:
@tannerlinsley Does opting out of this behavior for use with reach make sense and align with the direction your heading? I know you mentioned you were toying with a hook based router.

tannerlinsley:
I mostly wondering if we could get around this without adding any more api surface area, eg. find a core fix that just works invisibly. Can you point me to some code where you are experiencing the need for this?

georgehenderson:
Yeah ideally we shouldn't add more api surface area. Here is a codesandbox that demonstrates the problem through logs. https://codesandbox.io/s/o1oj410l5y .  Not sure why errors are being thrown in the sandbox but the code works if you pull it down.

georgehenderson:
The blog page and post page templates have logs that will double log on page change

tannerlinsley:
I believe this is now solved from #1015.

PR #992, opened 2019-01-29
Enable routeData for the 404 page

ashrewdmint:
## Description

Allow routeData to be used by the 404 page

## Changes/Tasks
- Edited `getRouteInfo()` to look for info on the `404` route if an error occurred trying to find info (unless it was trying to find info on the `404` route and failed)
- Edited `RouteData` component to check for info on the `404` route if a `routeError` was found for the current route, and if so, to use the 404 route info.

## Motivation and Context

In version 5, it was possible to pass `routeData` to the 404 route. However, in version 6 this is no longer possible. Although the 404 route may access `siteData`, this is not sufficient for all use cases. It is very convenient to be able to pass `siteData` to the 404 route in unexpected situations, including: having a localized header/footer and wanting the header/footer to appear in the 404 route, without complicated the existing codebase by, for example, putting English header/footer data into `siteData` (where it would add unnecessary weight to routes that don't need it, etc.).

## Types of changes

## Checklist:

### DIFF
--- a/packages/react-static/src/browser/components/RouteData.js
+++ b/packages/react-static/src/browser/components/RouteData.js
@@ -40,7 +40,7 @@ const RouteData = withStaticInfo(
 
         // If there was an error reported for this path, throw an error
         // unless there is data for the 404 page
-        if (routeError && !routeInfo) {
+        if (routeError && (!routeInfo || !routeInfo.data)) {
           throw new Error(
             `React-Static: <RouteData> could not find any data for this route: ${routePath}. If this is a dynamic route, please remove any reliance on RouteData or withRouteData from this routes components`
           )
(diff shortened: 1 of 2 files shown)


ashrewdmint:
I couldn't find any tests related to this part of the framework, but if I overlooked something let me know and I will add tests.

I'm also not sure if there might be unintended consequences to these changes, but I'm not sure what to look for because I'm not familiar with this codebase.

ashrewdmint:
```
if (routeError && !routeInfo) {
```

Hmm I wonder if this should be

```
if (routeError && (!routeInfo || !routeInfo.data)) {
```

sw-yx:
looks good - @tannerlinsley?

tannerlinsley:
Looks good :)

PR #1010, opened 2019-02-08
Fix loading of shared data

bram-l:
## Description

* Make sure shared data is always merged with data at the same point to create more predictable behaviour
* Load shared data when necessary
* Remove confusing return value of prefetch methods (since it's mutating the route info, the return value is never needed)

## Motivation and Context

Fix nozzle/react-static#981

I've added some helper functions to determine whether or not the route info and/or its shared data should be loaded client-side. Although I feel like these might not be in the right place, but I'm not sure what would be a more appropriate place. Any suggestions?

## Types of changes

## Checklist:

### DIFF
--- a/packages/react-static/src/browser/components/RouteData.js
+++ b/packages/react-static/src/browser/components/RouteData.js
@@ -46,10 +46,10 @@ const RouteData = withStaticInfo(
           )
         }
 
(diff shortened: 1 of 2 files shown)


tannerlinsley:
Excellent :)

PR #1012, opened 2019-02-09
Move build artifacts to artifacts dir (#1009)

### LINKED ISSUE #1009: [Bug] Build files making their way to dist folder
### Reporting a bug?

Hi there!

This could be a bug or a feature request depending on how you look at things. 

Running `react-static build` will not only add the assets from `public/` and the generated html, css, and js files, but it will also add several other files which should *not* be made public. 

To name a few: 
`client-stats.json`
`react-static-browser-plugins.js`
`react-static-build-config.json`
`react-static-templates.js`

It might be that these are build requirements but I don't see why they need to be dumped out on the dist folder and not on `tmp/`

This issue was briefly mentioned in an old issue I found https://github.com/nozzle/react-static/issues/664 but there was a regression at some point and the fix is long gone.

**Base your steps off of any freshly installed react-static template!**

1.  npm run build

### Expected Behavior

No sensitive information in dist folder

adrianObel:
Move build artifacts from `dist/` to a new, `artifacts/` directory.

## Changes/Tasks

- [x] Changed code

## Motivation and Context
Build artifacts with potentially private data were being grouped together with build files.
This change makes it so they are sent to `artifacts/` instead.

Closes #1009 

## Types of changes

## Checklist:

### DIFF
--- a/packages/react-static/src/commands/bundle.js
+++ b/packages/react-static/src/commands/bundle.js
@@ -55,6 +55,12 @@ export default (async function bundle({
   await fs.remove(config.paths.DIST)
   timeEnd(chalk.green('=> [\u2713] Dist cleaned'))
 
+  // Remove the ARTIFACTS folder
+  console.log('=> Cleaning artifacts...')
+  time(chalk.green('=> [\u2713] Artifacts cleaned'))
+  await fs.remove(config.paths.BUILD_ARTIFACTS)
+  timeEnd(chalk.green('=> [\u2713] Artifacts cleaned'))
+
   // Empty ASSETS folder
   if (config.paths.ASSETS && config.paths.ASSETS !== config.paths.DIST) {
     console.log('=> Cleaning assets...')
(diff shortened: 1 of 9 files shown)


adrianObel:
This change gets us closer but I'm not entirely happy. There are still some files which, with my limited understanding of the codebase, seem like they shouldn't be grouped together with build - especially production builds.

`static[hash].js`, and its mapping, look like dev dependencies for hot reloading or some other kind of bootstrapping. Could we skip out on these for production builds? They don't seem to be referenced in any of the build files.

How come the source maps generate a full path starting from `/`? I can't image that being very useful for prod builds where code isn't running on the same machine. I can sort of see if for `static[hash].js` if my last assumption is correct, but not in any of the other files that are produced.

I could very well be wrong about my assumptions so if there's a reason for things, I'm all ears (eyes?) 😄

tannerlinsley:
The static bundle it's only used for SSR, so that can mount into artifacts
as well.

bram-l:
Hi, I think this is a great idea! I was just pondering about the name 'artifacts'.. It sounds like those files are just a byproduct of the build, but if I understand correctly the are actually required if you want to support incremental builds. Perhaps something like 'staging' might be better in conveying the purpose of those files?

tannerlinsley:
I still vote artifacts I think. They are byproducts of the build, but not a lot of people are using incremental builds, so artifacts sounds like something you would normally gitignore. If they end up using incremental builds, they can remove it from the ignore list.

PR #1015, opened 2019-02-12
Fix remount of component on history update

### LINKED ISSUE #974: Tree remount on history.replaceState()
## Story

1. `history.replaceState()` is overridden by [Location.js#L26](https://github.com/nozzle/react-static/blob/v6.3.0/packages/react-static/src/browser/utils/Location.js#L26), this allows React Static to add own subscribers
2. Subscriber is added there, [Routes.js#L30](https://github.com/nozzle/react-static/blob/v6.3.0/packages/react-static/src/browser/components/Routes.js#L30), it will call `forceUpdate()`
3. `getComponentForPath()` will always return a new component, [Routes.js#L83](https://github.com/nozzle/react-static/blob/v6.3.0/packages/react-static/src/browser/components/Routes.js#L83):

```jsx
<Routes>
  {({ getComponentForPath, routePath }) => {
    const Component = getComponentForPath(routePath);
    console.log("Component", C === Component);
    console.log("Root", P === Root);
    C = Component;
    P = Root;
    return <Component key={routePath} />;
  }}
</Routes>
```

```jsx
// First render
Component false
Root false
// Fetch info about missing route
Component false
Root true
// After history.replaceState()
Component false
Root true
```

## Issue

The behavior described above works fine for a navigation between pages, but when I'm doing `history.replaceState(undefined, undefined, '#anchor')` it causes an update and remount of a whole tree. In my case it does a strange page jump.

```
componentDidMount()
// CLICK
history.replaceState()
componentDidMount()
```

static

### Environment

1.  `react-static -V`: 6.3.0

### Monkey resolution

```diff
-        return (newProps = {}) =>
-          Comp ? (
-            <Comp {...newProps} {...(is404 ? { is404: true } : {})} />
-          ) : null
+ return <Comp {...(is404 ? { is404: true } : {})} />
```

This works fine to me, it's too late there to find a better solution 🌃

bram-l:
## Changes
* Cache wrapped 404 component
* Prevent wrapping of non-404 components
* Remove function creation from render method

## Motivation and Context
Fixes #974

## Types of changes

## Checklist:

### DIFF
--- a/packages/react-static/src/browser/components/Routes.js
+++ b/packages/react-static/src/browser/components/Routes.js
@@ -20,6 +20,21 @@ export const withRoutePathContext = Comp => props => (
   </RoutePathContext.Consumer>
 )
 
+const componentCache = new WeakMap()
+function with404Prop(Component) {
+  // If the wrapped Component is currently in cached, return it from cache.
+  if (componentCache.has(Component)) {
+    return componentCache.get(Component)
+  }
+
+  // Otherwise, create a new wrapped Component...
+  const WrappedComponent = props => <Component is404 {...props} />
+
+  // ...and cache it
+  componentCache.set(Component, WrappedComponent)
+  return WrappedComponent
+}
+
 export default withStaticInfo(
   class Routes extends Component {
     static defaultProps = {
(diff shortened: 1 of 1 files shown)


TN1ck:
Just wanted to thank you guys for making this fix! I just run into this yesterday on 6.3.4 and was soo glad to have it fixed by simply upgrading this package. Thanks for your work ❤️

PR #1016, opened 2019-02-12
Fix sharedData fetching to use root-relative url

krzysztof-pyrzyk-ah:
## Description
Fetching static data was only working for `/` path, because it wasn't using root-relative url.
When opening a page with URL like f.ex `/blog/my-blog-post/`, which is using shared data, react-static is trying to fetch these data from `blog/my-blog-post/staticData/xxxxxxx.json`, but staticData folder is located at the root level of our page, so request should be prefixed with `/`.

## Changes/Tasks
- [x] Moved staticDataPath to a separate variable
- [x] Fixed fetching sharedData by using an URL with leading slash
- [x] Removed unnecessary space in PR's template

## Types of changes

## Checklist:


bram-l:
@tannerlinsley Did you have a look at this? I've been testing this thoroughly on my locale setup and it seems this would be the last fix to resolve the shared data issues (which is currently a blocker for us)

krzysztof-pyrzyk-ah:
@bram-l Thanks for testing my pull request. For sure it fixes some issues, but honestly I still struggle with sharedData on `build` command.

bram-l:
@krzysztof-pyrzyk-ah Did you try using the master branch (which includes #1019)?

krzysztof-pyrzyk-ah:
@bram-l This may be the case! I've been using newest tag, as I thought your PR was released. I'll try that. Thanks ;)

krzysztof-pyrzyk-ah:
@tannerlinsley Can I please ask for a Code Review? It's been a long time since this PR was created and the problem it solves hasn't been fixed anywhere else.

tannerlinsley:
So what would happen if the asset path was set by the user to something like `https://cdn.com/mybucket`? Wouldn't this produce an invalid link? eg. `/https://cdn.com/mybucket`?

krzysztof-pyrzyk-ah:
@tannerlinsley Good point, it would produce an invalid link. Haven't thought about that.
I'll have to come up with a new solution.
I'll take a look at `pathJoin` method cause asset path is by default set to `/` which is fine, but `pathJoin` method cleans leading slash by default.

bram-l:
Yes that's right, I also didn't think about that... It looks like `makePathAbsolute` might come handy here!

tannerlinsley:
Nice work!

PR #1020, opened 2019-02-13
Log threads not cores from export

rupurt:
## Description
Log thread count instead of core count

## Changes/Tasks

- [ ] Changed code

## Motivation and Context
It was incorrectly logging cores instead of threads

## Types of changes


tannerlinsley:
Nice catch!

PR #1024, opened 2019-02-16
Hard-cache babel config using `api.cache(true)`

### LINKED ISSUE #1023: [Bug] Babel, "Caching was left unconfigured."
I just setup a new install and updated my packages. Now I'm getting the following error:

```
  Error: Caching was left unconfigured. Babel's plugins, presets, and .babelrc.j  s files can be configured
  for various types of caching, using the first param of their handler functions  :
  module.exports = function(api) {
    // The API exposes the following:
    // Cache the returned value forever and don't call this function again.
    api.cache(true);
    // Don't cache at all. Not recommended because it will be very slow.
    api.cache(false);
    // Cached based on the value of some function. If this function returns a va  lue different from
    // a previously-encountered value, the plugins will re-evaluate.
    var env = api.cache(() => process.env.NODE_ENV);
    // If testing for a specific env, we recommend specifics to avoid instantiat  ing a plugin for
    // any possible NODE_ENV value that might come up during plugin execution.
    var isProd = api.cache(() => process.env.NODE_ENV === "production");
    // .cache(fn) will perform a linear search though instances to find the matc  hing plugin based
    // based on previous instantiated plugins. If you want to recreate the plugi  n and discard the
    // previous instance whenever something changes, you may use:
    var isProd = api.cache.invalidate(() => process.env.NODE_ENV === "production  ");
    // Note, we also expose the following more-verbose versions of the above exa  mples:
    api.cache.forever(); // api.cache(true)
    api.cache.never();   // api.cache(false)
    api.cache.using(fn); // api.cache(fn)
    // Return the value that will be cached.
    return { };
  };
```

There's [an issue](https://github.com/babel/babel/issues/9104) for it already, but it's not really a bug. It might be worth actually using `api.cache()`.

mikestopcontinues:
## Description
Given that we only continuously compile during during dev, and given that the config only depends on `external` which isn't likely to change during development, a hard cache seems like the right solution.

## Motivation and Context
Fixes #1023

## Types of changes

## Checklist:

### DIFF
--- a/packages/react-static/babel-preset.js
+++ b/packages/react-static/babel-preset.js
@@ -4,6 +4,8 @@ module.exports = (api, { external, modules, helpers } = {}) => {
   const { NODE_ENV, BABEL_ENV } = process.env
 
   const PRODUCTION = (BABEL_ENV || NODE_ENV) === 'production'
+  
+  api.cache(true)
 
   return {
     ...(external


tannerlinsley:
Good idea

PR #1029, opened 2019-02-18
Improve prefetchable route detection

bram-l:
## Description
* Don't rely on 'invalid' relative links, but check for hostname, protocol and port instead, fixing detection of absolute URLs without explicit protocol
* Removed redundant check for script links
* Added unit tests

## Motivation and Context
`//www.example.com` would be considered 'prefetchable' because it would throw an error, and all invalid URLs were considered 'prefetchable'. By adding the base URL this will be fixed, it will create a valid URL object, but the hostname will differ and therefor the link will not be considered 'prefetchable'.

## Types of changes

## Checklist:


tannerlinsley:
This looks fantastic! Great work.

PR #1030, opened 2019-02-18
Restore route data return value for prefetch

bram-l:
## Motivation and Context
Previously, in #1010 I had removed the return value of the `prefetch` method, since it was not used internally and seemed only confusing. Now I just realised this was actually a documented feature 😨 ... I restored the return value, making sure only for routes resolving in a 404  it will return `undefined`.

## Types of changes

## Checklist:


tannerlinsley:
I would have been okay removing the docs but this is good too 😉. Nice work!

tannerlinsley:
Can you pull in the latest commits and unconflict?

PR #1034, opened 2019-02-19
Move `eslint-config-react-tools` to devDeps

### LINKED ISSUE #1033: [Bug] `eslint-config-react-tools` dependency messes up local eslint config.
### Reporting a bug?

I use a third-party eslint-config. The trouble is, `react-static` installs `eslint-config-react-tools` as a standard dependency. Node seems to give priority to this package's eslint-plugins, which causes some errors, as the version of `eslint-plugin-babel` it requires is pretty old.

```
  1:1  warning  Definition for rule 'babel/quotes' was not found                 babel/quotes
  1:1  warning  Definition for rule 'babel/no-unused-expressions' was not found  babel/no-unused-expressions
  1:1  warning  Definition for rule 'babel/valid-typeof' was not found           babel/valid-typeof
```

It doesn't look like the dep is being used for anything at all, but I'm attaching a PR where `eslint-config-react-tools` is simply moved to `devDependencies`. If nothing else, it cuts the `react-static` down by a big chunk.

mikestopcontinues:
## Description
Fixes #1033. 

The goal here is to prevent conflicts with other third-party eslint-configs and reduce install size.

## Changes/Tasks
- [x] Changed deps

## Types of changes

## Checklist:


tannerlinsley:
Good call

PR #1036, opened 2019-02-21
Add synchronous getWebpackConfig method to node api.

jaridmargolin:
## Description
Added synchronous `getWebpackConfig` method to node api.

## Changes/Tasks

- [X] Changed code

## Motivation and Context
I believe there are several benefits to providing this method, but it came to light as I attempted to get `eslint` working correctly with `react-static`.

The problem is that we use [eslint-import-resolver-webpack](https://www.npmjs.com/package/eslint-import-resolver-webpack) in order for eslint to be capable of resolving dependencies the same way in which webpack would. The only way to know how webpack would resolve, is to expose a `webpack.config.js` to `eslint-import-resolver-webpack`.

In my first attempts, I utilized the existing internal methods and took advantage of the fact that webpack allows exporting a promise: https://webpack.js.org/configuration/configuration-types/#exporting-a-promise.

Unfortunately, eslint requires running synchronously: https://github.com/benmosher/eslint-plugin-import/issues/883.

As I dug a little deeper into the codebase of react-static, I didn't see a compelling reason why the internal methods I was using were async.

This unfortunately is a breaking change, as the `config` and `webpack` hooks now require resolving synchronously. All documentation, and all implementations I have seen, currently resolve synchronously, so hopefully the collateral damage will be small.

I believe this change is a net win. Yes we are removing some functionality, but we are doing so in order to support a much more common use case.

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

### DIFF
--- a/packages/react-static/node/index.js
+++ b/packages/react-static/node/index.js
@@ -5,6 +5,7 @@ const { default: bundle } = require('../lib/commands/bundle')
 const { default: exporter } = require('../lib/commands/export')
 const { reloadRoutes } = require('../lib/static/webpack')
 const { default: makePageRoutes } = require('../lib/node/makePageRoutes')
+const { default: getWebpackConfig } = require('../lib/node/getWebpackConfig')
 const { normalizeRoutes } = require('../lib/static/getConfig')
 const { default: createSharedData } = require('../lib/static/createSharedData')
 
@@ -16,6 +17,7 @@ module.exports = {
   export: exporter,
   reloadRoutes,
   makePageRoutes,
+  getWebpackConfig,
   normalizeRoutes,
   createSharedData,
 }
--- a/packages/react-static/src/node/getWebpackConfig.js
+++ b/packages/react-static/src/node/getWebpackConfig.js
@@ -1,0 +1,9 @@
+import getConfig from '../static/getConfig'
+import { webpackConfig } from '../static/webpack'
+
+// Required so handle static.config.js defined as es module
+require('../utils/binHelper')
+
+export default (function getWebpackConfig(configPath, stage = 'dev') {
+  return webpackConfig({ config: getConfig(), stage })
+})
--- a/packages/react-static/src/static/getConfig.js
+++ b/packages/react-static/src/static/getConfig.js
@@ -23,7 +23,7 @@ const DEFAULT_ROUTES = [{ path: '/' }]
 const DEFAULT_ENTRY = 'index'
 const DEFAULT_EXTENSIONS = ['.js', '.jsx']
 
-export const buildConfig = async (config = {}) => {
+export const buildConfig = (config = {}) => {
   // path defaults
   config.paths = {
     root: nodePath.resolve(process.cwd()),
@@ -259,20 +259,18 @@ export const buildConfig = async (config = {}) => {
 
   config.plugins = config.plugins.map(resolvePlugin)
 
-  const configHook = makeHookReducer(config.plugins, 'config')
-  config = await configHook(config)
-
-  return config
+  const configHook = makeHookReducer(config.plugins, 'config', { sync: true })
+  return configHook(config)
 }
 
-const buildConfigFromPath = async configPath => {
+const buildConfigFromPath = configPath => {
   delete require.cache[configPath]
   const config = require(configPath).default
   return buildConfig(config)
 }
 
 // Retrieves the static.config.js from the current project directory
-export default (async function getConfig(
+export default (function getConfig(
   configPath = DEFAULT_PATH_FOR_STATIC_CONFIG,
   subscription
 ) {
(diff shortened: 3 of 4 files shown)


tannerlinsley:
My gut tells me that we could still achieve this somehow without these changes, but I'm not sure on the details yet. I feel like it might involved pointing eslint to to a dummy webpack config, then decorating that config as react-static progressively boots up.

I'll keep thinking about this.

jaridmargolin:
@tannerlinsley - Sure. I just unfortunately wasn't able to devise anything that seemed like a better solution.

I did have one thought that I could build the webpack config asynchronously and then programmatically call eslint. While in theory this would allow me to run eslint manually, my editor plugin still wouldn't be capable of working correctly.

I concluded that building synchronously would be the most expected behavior for my use-case.

Look forward to seeing if you can come up with anything :) Thanks!

tannerlinsley:
After thinking about this more. I don't feel good about turning the api for this synchronous, but I also don't want what you're talking about to be impossible. Going asynchronous out of the gate was a flexibility play and there are some use cases where people are using external asynchronous functions to manipulate the config. Whether those funcs should be async or not is another question, but nonetheless, it's the least common denominator, so we need to support it. 

Although this isn't the response I wanted to give, I think it's much easier for us to request/contribute to upstream packages to support async+sync solutions, rather than force out half of the flexibility of our own API.

With that in mind, what would need to happen upstream to make this work?

jaridmargolin:
@tannerlinsley - I understand not wanting to remove functionality, but I do want to ask you to reconsider.

`react-static` v6 does not currently support `eslint`, one of the most used tools in the js eco-system, accounting for over 5 million downloads a week. The trade-off for not supporting `eslint`, is to support the edge case of needing to resolve `config`/`webpack` asynchronously (which isn't [currently documented](https://github.com/nozzle/react-static/blob/master/docs/plugins/node-api.md#config-function)). Do we have concrete examples of where this is necessary?

Pushing this upstream would require changes to `eslint` core to expose asynchronous hooks in its plugin architecture. It would then require `eslint-plugin-import` to utilize these new hooks. I'm not saying there isn't a valid use case for this functionality in `eslint`, but I think accomplishing this would be a monumental task.

Is there anything else I could say to persuade you here?

Thanks for your time looking into this.

tannerlinsley:
I think generalizing this issue as RS not supporting eslint is incorrect. It's definitely widely supported in general, just not this `eslint-import-resolver-webpack` extension, correct? 

That being said, I can see your point. I think the best way forward is going to ask you to wait until v7 when we can implement this as a breaking change.

jaridmargolin:
@tannerlinsley Hopefully I'm not coming off as confrontational. I was not attempting to make a broad generalization, but I do see how it could be interpreted that way.

The way I see the problem (which potentially is different than the way you see the problem), is that `react-static` leverages `webpack`. In turn, `eslint-import-resolver-webpack` is required to accurately run `eslint`.  Because I cannot accurately run `eslint`, I would say that `react-static` does not support `eslint`.

If you are willing to change the behavior, which sounds like you are still considering, I imagine v7 is not in the near future (as v6 was just released on Dec 12, 2018). How are releases managed?

tannerlinsley:
v7 will likely come in a few months. RS moves pretty fast. There's nothing sacred about a version number, we just honor semver. With every release, things change, patterns mature, upstream things get better, so usually there are a few ongoing issues to track slated breaking changes (like this one). 

There is a v7 branch right now that has a few commits, but is definitely broken as is. We're still working on the exact specs for it (don't even have a readme/todo list yet, but will soon).

jaridmargolin:
@tannerlinsley - I have a proposal. This is currently blocking us from upgrading to react-static v6 😭and I'd prefer not to manage a separate fork.

1. I expose `getWebpackConfig` and `getWebpackConfigSync` methods.
2. I alter the internals to provide an optional synchronous code path.
3. If the sync route is followed and a promise is returned in any of the config/webhook methods, an error is thrown notifying the user that the config can not be resolved synchronously.

Benefits of this approach:
1. It allows us to shop this behavior without having to wait for v7 🎉.
2. It is completely additive.
3. The error handling protects against any unexpected behavior.
4. This does not alter how `react-static` will run. The synchronous code path with only ever be followed when `getWebpackConfigSync` is called.

Downsides:
1. Additional internal complexity.
2. Slightly expanded API.

tannerlinsley:
Based on your last message, is this PR ready to merge?

jaridmargolin:
@tannerlinsley - Thanks for getting back to me. No, I have not updated the PR. I wanted to confirm you were onboard with the premise before I implemented.

tannerlinsley:
Yep, let's do it. Keep it minimal, and make sure things are fairly future proof and portable as there could be some changes coming with v7.

jaridmargolin:
Closing in favor of #1064

PR #1037, opened 2019-02-22
Add custom data argument for route rebuilds

dpikt:
Adds a method for passing in custom data to `getRoutes()` in the node API commands. If this approach is approved, I can add docs and tests.

## Description

An example of passing custom data:
```
// Rebuilding from webhook (server.js)
reloadRoutes(null, { data: 'some token from webhook' })

...

// In static config
getRoutes: async ({ data: token }) => {
    const data = await Prismic.fetch({ token })
    return routes(data) // return routes array
},
```

## Changes/Tasks
- [x] Changed code
- [ ] Update docs
- [ ] Update tests

## Motivation and Context
This is useful for when you need data to change for each given rebuild. For example, Prismic live site previews that need to fetch data from a certain ref.

## Types of changes

## Checklist:

### DIFF
--- a/packages/react-static/src/commands/bundle.js
+++ b/packages/react-static/src/commands/bundle.js
@@ -17,6 +17,7 @@ export default (async function bundle({
   config: originalConfig,
   staging,
   debug,
+  data,
 } = {}) {
   // ensure ENV variables are set
   if (typeof process.env.NODE_ENV === 'undefined' && !debug) {
@@ -70,7 +71,7 @@ export default (async function bundle({
   }
 
   config = await prepareBrowserPlugins(config)
-  config = await prepareRoutes(config)
+  config = await prepareRoutes(config, { data })
   await extractTemplates(config)
   await generateTemplates(config)
 
--- a/packages/react-static/src/commands/start.js
+++ b/packages/react-static/src/commands/start.js
@@ -15,7 +15,7 @@ import { createIndexFilePlaceholder } from '../utils'
 let cleaned
 let indexCreated
 
-export default (async function start({ config: configPath, debug } = {}) {
+export default (async function start({ config: configPath, debug, data } = {}) {
   // ensure ENV variables are set
   if (typeof process.env.NODE_ENV === 'undefined') {
     process.env.NODE_ENV = 'development'
@@ -52,7 +52,7 @@ export default (async function start({ config: configPath, debug } = {}) {
 
     config = await prepareBrowserPlugins(config)
 
-    await prepareRoutes(config, { dev: true }, async config => {
+    await prepareRoutes(config, { dev: true, data }, async config => {
       await extractTemplates(config, { dev: true })
       await generateTemplates(config)
       reloadRoutes()
--- a/packages/react-static/src/static/webpack/index.js
+++ b/packages/react-static/src/static/webpack/index.js
@@ -246,7 +246,7 @@ export async function startDevServer({ config }) {
   // Start the messages socket
   const socket = io()
 
-  resolvedReloadRoutes = async (paths, data) =>
+  resolvedReloadRoutes = async (paths, { data } = {}) =>
     prepareRoutes(config, { dev: true, silent: true, data }, async config => {
       if (!paths) {
         paths = config.routes.map(route => route.path)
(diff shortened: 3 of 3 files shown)


tannerlinsley:
My gut reaction is that this can be avoided via some added user code in the config. Maybe something like:
```
let prismaRef

prisma.subscribe(ref => {
  prismaRef = ref
  reloadRoutes()
})

export default {
  getRoutes: () => {
    // use prismaRef?
  }
}

**Comment by dpikt** [comment]
Thanks for the quick response! That's definitely a viable solution. My hangups with that are that it requires everything to live in the same file (unless you used some sort of store like redis) and it's less "pure" in terms of inputs and outputs (getRoutes would be relying on external state which might be more difficult to test/debug).

The first point is more important to me since I have an app that contains a static site and a dynamic "preview" site and it's nice to break things out:

```
// Preview app runs dev build since routes need to change

import { start, reloadRoutes } from 'react-static/node'
import baseConfig from '../static.config'
import { set } from 'lodash/fp'

function previewMiddleware (app) {
  return app.get('/preview', async (req, res) => {
    await reloadRoutes(null, { data: req.query.token })
    res.send('Preview built')
  })
}

const config = set('devServer.before', previewMiddleware, baseConfig)

start({ config })

```

Let me know if this use case makes sense to you, or if you think this is a break from the standard mental model of react static that shouldn't be replicated.

**Comment by tannerlinsley** [comment]
I might be able to get on board. Most of my hesitation simply stems from expanding the API in general. In most scenarios, I'm trying to cut down on the API size while making it more flexible/composable. 

Let's keep talking about this and see where it goes.

**Comment by dpikt** [comment]
Ok, I've done some soul-searching and I think I've changed my mind.

If I think about using the Node API like the webpack CLI, then the work really should be done in the config as you suggested.

That being said, it's still possible to split out the files by extending the config directly:

```
// Preview app runs dev build since routes need to change

import { start, reloadRoutes } from 'react-static/node'
import baseConfig from './static.config'
import { compose, set } from 'lodash/fp'

let prismicRef = null

function previewMiddleware (app) {
  return app.get('/preview', async (req, res) => {
    prismicRef = req.query.token
    await reloadRoutes()
    res.send('Preview built')
  })
}

const config = compose(
  set('devServer.before', previewMiddleware),
  set('getRoutes', () => baseConfig.getRoutes(prismicRef)),
)(baseConfig)

start({ config })
```

It's a little muddy because this "rebuilding" concept (feeding back into a compiled app) doesn't show up with the typical webpack dev server setup, but I'm satisfied with this solution.

One note- it looks like support for passing in a configuration object directly was dropped in v6. Would you be alright with adding it back? If so, I can add a PR for that.

dpikt:
@tannerlinsley Is passing a config object to `start()` directly something you still want to support?

tannerlinsley:
If I recall correctly, the node api for starting, building, exporting ect was removed a while back no?

dpikt:
I don't think so... at the very least it's still documented [here](https://github.com/nozzle/react-static/blob/master/docs/node-api.md) and importable.

tannerlinsley:
Oh you're totally right! My bad.

So this is one of those things that I've been going back and forth on. I personally haven't liked the pattern in the past of providing 2 ways to do the same thing and this is one of those things. Unless it provides a really big value add, anything should be considered for deprecation.

Specifically for the start, create, build, bundle, and export methods, I don't see them offering much convenience, especially when that means maintaining all of the edge cases that come with it. Notably, in the past, we've had problems where people expect the same babel support, plugins, and environment using `start()` that they do with `react-static start` and they're just very different entry points.

What are your thoughts?

dpikt:
Makes sense to me. My main mental model for this build tool is webpack, and I've definitely noticed a trend away from using the node API as the configuration object becomes more fully featured. That being said, I think having the option to start compilations programmatically is still useful, as long as those compilations behave the same way as the CLI. Perhaps a good next step would be to iron out the differences between entry points to minimize further issues? I haven't dug into that code much so I'm not sure how feasible that is.

In the meantime, would you like me to update the docs to reflect that only a configuration path is accepted?

tannerlinsley:
Yes. I think that's a good idea regardless.

dpikt:
Cool, I'll open a PR for that change. Thanks!

PR #1041, opened 2019-02-22
Move node app build to artifacts folder

dpikt:
## Description
Moves the output of the "node" app build to the artifacts folder. I also added a change that ignores file size limits to the URL loader, but was considering ignoring these files entirely (curious about your thoughts on this).

## Changes/Tasks
- [x] Changed code

## Motivation and Context
Since this compiled app is only used by route exporters, moving it out of the build folder should make its purpose more explicit while saving space.

## Types of changes
- [x] Refactoring/add tests (refactoring or adding test which isn't a fix or add a feature)

## Checklist:


tannerlinsley:
Love this. Good move.

PR #1054, opened 2019-03-05
Fix shared route data for 404 pages

bram-l:
## Description
When fetching shared data for 404 pages we should use the path to the 404 instead of the original path

## Motivation and Context
Currently react-static will fail to load the shared data for 404 pages when navigating from another page, and it will simply remain in a loading state.

## Types of changes

## Checklist:


krzysztof-pyrzyk-ah:
It would be super nice if we could get a new version tag released along with these changes 🙏

PR #1055, opened 2019-03-05
Include port in base url when matching templates

dpikt:
Includes `port` in the base URL definition when template matching. 

## Description

Fixes a bug where routes weren't being matched correctly when using `localhost` (all routes returning 404). This was occurring because the port was being dropped from the base url definition.

## Changes/Tasks

- [x] Changed code

## Types of changes

## Checklist:


tannerlinsley:
Nice work!

PR #1067, opened 2019-03-13
prevent source code provisioning in release build

neilyoung:
Currently source code is deployed to the server even after `yarn build`

## Description
changed some flags in webpack.config.prod.js

### DIFF
--- a/packages/react-static/src/static/webpack/webpack.config.prod.js
+++ b/packages/react-static/src/static/webpack/webpack.config.prod.js
@@ -125,7 +125,7 @@ function common(config) {
       new CaseSensitivePathsPlugin(),
       config.bundleAnalyzer && new BundleAnalyzerPlugin(),
     ].filter(d => d),
-    devtool: undefined,
+    devtool: false,
   }
 }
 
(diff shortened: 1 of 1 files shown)


packages/react-static/src/static/webpack/webpack.config.prod.js
```diff
@@ -125,7 +125,7 @@ function common(config) {
       new CaseSensitivePathsPlugin(),
       config.bundleAnalyzer && new BundleAnalyzerPlugin(),
     ].filter(d => d),
-    devtool: 'source-map',
+    devtool: undefined,
   }
```
tannerlinsley: Why not `false`?

https://webpack.js.org/configuration/devtool#devtool

neilyoung: Wasn't sure. Change it, if you like. Undefined does work too

neilyoung: I thought it defines a command, so false for a command - well, great JS world :)

tannerlinsley: While it does work in this scenario, `undefined` can sometimes denote the intent of "use the default". Let's go with `false`.

neilyoung:
Thanks

Issue #985, opened 2019-01-22
[Bug] 404s every page after enabling watch('content') to reloadRoutes() in static.config.js


theWebalyst:
### Are you reporting a bug?

Yes

Note: this issue originally filed on my rs-basic-test repo: https://github.com/theWebalyst/rs-basic-test/issues/1

### Environment

    react-static -V: 6.1.5
    node -v: 8.11.3
    npm -v: 6.4.1
    Operating system: Ubuntu 18.10

### Steps to Reproduce the problem

Starting with the basic template, I made small changes to get it to load content from markdown files by changing the relevant parts, based this other example: https://github.com/tannerlinsley/react-static-markdown (I chose to follow how that loads markdown having first failed to do this based on the code in the react-static archived markdown example).

You can reproduce the error by:

- `git clone https://github.com/theWebalyst/rs-basic-test/ && cd rs-basic-test && git checkout tags/take-content-from-markdown-files -b localtest && yarn && yarn start`
- visit http://localhost:3000/ and confirm Home, About, Blog etc load ok
- remove the comment in static.config.js to enable reloadRoutes() at [line 7](https://github.com/theWebalyst/rs-basic-test/blob/take-content-from-markdown-files/static.config.js#L7)
- things work as expected if you *just edit content*
- until you refresh the page or click on the navigation: Home, About, Blog etc all show a 404 error

I don't see any errors in browser or terminal console (using Firefox), but note that there is still some console.log() debug output generated to the terminal by the build (see static.config.js 15-17)

gpoole:
I'm having the same issue, but I don't think it's caused by the call to `reloadRoutes()`. If I make any changes at all to my static.config.js and save I get 404s until I restart the watch server. I've also checked and the same thing happens in the linked example without the `reloadRoutes()` line being uncommented.

theWebalyst:
Phew, it isn't just me - always the expectation when learning something new :smile: 

I just updated the OP to reflect that if I start without the `reloadRoutes()` line, do ỳarn start` reload the site (all good), edit `static.config.js` to enable:
```
chokidar.watch('content').on('all', () => reloadRoutes())
```
Then save `static.config.js` it all works. Changes to my content are hot reloaded as expected... *until* I refresh the page, or click on any navigation, or restart `yarn start`. Then it is all 404s again.

I'm not sure if this is consistent with your report @gpoole?

gpoole:
[...]

PR #1075, opened 2019-03-22
Amend examples and docs to reflect that chokidar watcher needs to be closed for build to finish

lostfictions:
## Description

I discovered -- after much confusion -- that if you have a persistent chokidar watcher set up in your `static.config.js` and you try to run `react-static build`, the build will successfully complete, but then hang forever. This PR amends the examples and docs to reflect that the watcher needs to be closed in `onBuild`.

Doesn't look like this repo has any integration tests, otherwise I would've added one for verifying that a build that uses chokidar doesn't time out.

It looks like at least one or two of these examples are bit by #985, so this won't fundamentally make them work -- changing any content will still completely break react-static and will require a server restart to make things work again. But at least builds will work on CI (and won't require being ctrl-c'd out of locally). fwiw #985 seems like a serious, critical bug and I'm not sure react-static is usable in a lot of scenarios until it's fixed.

On another note, I'm not sure why the examples were all "deprecated" in one fell swoop -- many of them contain useful info that still isn't available in guides. I saw the explanation in the docs about why this happened, but I also think that examples and guides serve complementary usecases; some people find it easier to get started from one or the other. ([next.js](https://github.com/zeit/next.js/tree/canary/examples)'s examples directory might seem a little extreme, but I really appreciate that they put work in to keep all of them up-to-date and to explain the motivation behind each one.)

## Changes/Tasks
- [x] Changed code
- [x] Added comment to usage of chokidar in node-api.md

## Types of changes

## Checklist:

### DIFF
--- a/archives/old-examples/documentation/package.json
+++ b/archives/old-examples/documentation/package.json
@@ -11,6 +11,8 @@
   },
   "dependencies": {
     "axios": "^0.18.0",
+    "chokidar": "2.0.4",
+    "fs-extra": "^7.0.1",
     "react": "^16.6.3",
     "react-click-outside": "^3.0.1",
     "react-dom": "^16.6.3",
--- a/archives/old-examples/documentation/static.config.js
+++ b/archives/old-examples/documentation/static.config.js
@@ -5,17 +5,15 @@ import React, { Component } from 'react'
 import { ServerStyleSheet } from 'styled-components'
 import chokidar from 'chokidar'
 
-chokidar.watch('../docs').on('all', () => reloadRoutes())
-
-//
-
 const packageFile = 'package.json' // Point this to your package.json file
 const repoName = 'Blazing Awesome' // Change this
 const repo = 'nozzle/react-static' // Change this
 const repoURL = `https://github.com/${repo}`
 
+let watcher
+
 try {
-  // eslint-disable-next-line
+  // eslint-disable-next-line import/no-dynamic-require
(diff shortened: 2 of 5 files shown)


tannerlinsley:
You may want to check out [the changes being done in v7 for loading and watching routes from a directory](https://github.com/nozzle/react-static/tree/v7/packages/react-static-plugin-source-directory) before we move forward.

- The issue may already be resolved with those changes in v7, or your PR might be able to improve these changes
- I am *so happy* you figured out that this is why things are hanging.  It makes a ton of sense now!
- The examples directory was nice for some things, but honestly it became a logistical problem. Unfortunately I am still the only very active maintainer of RS and it was simply too much technical debt. I still have them saved in the repo in https://github.com/nozzle/react-static/tree/v7/archives/old-examples. This has helped keep things maintainable for me, but I see your point.

lostfictions:
Thanks for the reply! I took a look at the v7 tree -- all those changes sound great. If the new source-directory plugin is just for loading and watching routes, I don't think it impacts this PR -- these changes are for the scenarios where a directory of markdown files gets watched so that their data can be passed down to routes, regardless of where or how they're defined. Assuming I'm not misunderstanding the use of the plugin?

tannerlinsley:
Yeah, this shouldn't affect v7 or any new stuff. If you could, fix up this PR with master again so we can get it merged without conflicts?

PR #1076, opened 2019-03-24
makes sure animated example works correctly with the default App.tsx

capaj:
## Description
the latest react-spring does not have the Transition component API and also Transition duplicates the html it is applied on, and this does create weird jumping of the content unless `position: absolute` is applied

## Motivation and Context
Make the example work correctly with the default html structure

### DIFF
--- a/docs/guides/animated-routes.md
+++ b/docs/guides/animated-routes.md
@@ -2,7 +2,7 @@
 
 Animated Routes can be achieved so many different ways. In this example, we'll stick to something simple and use the `react-spring` package.
 
-- Install the `react-spring` module using npm or yarn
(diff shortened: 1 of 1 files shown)


tannerlinsley:
This guide may be outdated for v7. Could you please test with v7 and make sure?

capaj:
I did. Works well with version 7.

tannerlinsley:
Actually, after a closer look, this should be working since the Routes child renderer has been deprecated.

I'm not sure on the exact example yet, but we'll need to update it to use the new `Routes` component together with the `useRoutePath`. It's a bit different, so I'll work on an example to show you.

PR #1078, opened 2019-03-26
fix corejs warning for @babel/preset-env

digitalkaoz:
see https://github.com/gatsbyjs/gatsby/pull/12781/files


digitalkaoz:
ah it seems its only needed for a newer `preset-env` version...

tannerlinsley:
Please check out the latest v7. Is this still needed?

PR #1090, opened 2019-04-03
Fix for broken reloadClientData()

Blobson:
## Description
In the original version importing `reloadClientData` inside **static.config.js** end up with an empty function declared [here](https://github.com/nozzle/react-static/blob/daf2d6bd17c3365d22beee086d0fb79e7431ec91/packages/react-static/src/static/webpack/runDevServer.js#L15) which is substituted [here](https://github.com/nozzle/react-static/blob/daf2d6bd17c3365d22beee086d0fb79e7431ec91/packages/react-static/src/static/webpack/runDevServer.js#L217) after socket creation, but substituted version can't be propagated to imports made before the substitution. 
Fixed version exports a wrapper function which will call up-to-date version of `reloadClientData()`.

The second part of the fix is dedicated to invalidating `siteData` inside `onReloadClientData()` handler of `useSiteData()` hook. This ensures that `siteData` is always reloaded on `reloadClientData()` call.

The pull-request is only a proof of concept. I will be surprised if it does not break anything.

## Changes/Tasks
- [x] Changed code

## Motivation and Context
Fix reloading siteData changes via `reloadClientData()`

## Types of changes

## Checklist:

### DIFF
--- a/packages/react-static/src/browser/hooks/useSiteData.js
+++ b/packages/react-static/src/browser/hooks/useSiteData.js
@@ -16,6 +16,8 @@ export const useSiteData = () => {
   const [_, setCount] = useState(0)
   useEffect(() =>
     onReloadClientData(() => {
+      siteDataPromise = null
+      siteDataReady = false
       setCount(old => old + 1)
     })
   )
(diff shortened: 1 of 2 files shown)


tannerlinsley:
I think you've uncovered a pattern that needs to be fixed. I'm going to accept this PR and do some additional work. Thanks for finding this!

PR #1104, opened 2019-04-08
chore: fix types

cwaring:
I am developing a typescript react-static project and ran into the following issue while upgrading to v7 and using `useSiteData()` hook, which is blocking the build from compiling.

```
TS2349: Cannot invoke an expression whose type lacks a call signature. Type '{}' has no compatible call signatures.
``` 

## Description

I've changed `useSiteData` to the same signature as `useRouteData` which resolves the issue above. Additionally simplified the `react-helmet` types to a passive export.

## Changes/Tasks

- [x] Changed `useSiteData` type signature to fix the error
- [x] Removed the additional `Head` class and exported `{ Helmet as Head }` directly from `react-helmet`

## Motivation and Context

Fixes the error above that is currently blocking the app from compiling with `react-static` v7.0.6

## Types of changes

## Checklist:

### DIFF
--- a/packages/react-static/src/index.d.ts
+++ b/packages/react-static/src/index.d.ts
@@ -15,9 +15,8 @@ declare module 'react-static' {
   type AnyReactComponent = ComponentType<Record<string, any>>
 
   // Passing on helmet typings as "Head"
-  import { Helmet } from 'react-helmet'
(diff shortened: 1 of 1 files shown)


sw-yx:
thanks for your help on this!!

PR #1106, opened 2019-04-09
Corrected useSiteData() definition

Reggino:


Reggino:
Closing since already part of #1104

PR #1113, opened 2019-04-11
Fix using custom config with CLI

### LINKED ISSUE #1111: [Bug] Custom config file is not being used
It looks like the custom config path is not being used. Tracing the code, it looks like:

1. You run `react-static build -c ./custom.config.js`
2. config is being stored in as `config` (config.config) and passed into the invoked command: https://github.com/nozzle/react-static/blob/master/packages/react-static/bin/react-static#L22
3. and subsequently passed into `bundle` as `state` (so the path is now at `state.config`) in `build` command: https://github.com/nozzle/react-static/blob/master/packages/react-static/src/commands/build.js
4. which again got passed into `getConfig`: https://github.com/nozzle/react-static/blob/master/packages/react-static/src/commands/bundle.js#L29
5. `getConfig` tries to read it from `configPath` which is never defined: https://github.com/nozzle/react-static/blob/master/packages/react-static/src/static/getConfig.js#L33

### Environment

react-static 7.0.7

1.  create a custom config file and remove the default one
2.  react-static build -c custom.config.js

### Expected Behavior

1. It should use the custom config file but it would complain about not being able to find the default config file.

### Reproducible Demo

I feels like this is unnecessary as we found out where the error is already. I'm happy to put together a demo if we feel that's necessary. :)

Thanks!

itszero:
Passing a `--config` argument to the CLI does not work in v7 due to a key mismatch between CLI bootstrapping and getConfig function.

## Description

Changed the key for config file path from `config` to `configPath` to match the use case in `getConfig.js`

## Motivation and Context

This fixes #1111

## Types of changes

## Checklist:

## Notes:

This is my first contribution to react-static. Let me know how can I do better! :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 7.0.8
+
+### Improved
+
+- Fixed issues with custom config specified using `-c` or `-config` is not loaded
+
 # 7.0.7
 
 ### Improved
--- a/packages/react-static/bin/react-static
+++ b/packages/react-static/bin/react-static
@@ -19,7 +19,7 @@ function init() {
   const help = argv.help || argv.h || argv.H
   const options = {
     version: argv.version || argv.v || argv.V,
-    config: argv.config || argv.c || argv.C,
+    configPath: argv.config || argv.c || argv.C,
     name: argv.name || argv.n || argv.N,
     template: argv.template || argv.t || argv.T,
     staging: argv.staging || argv.s || argv.S,


tannerlinsley:
This is great! Let's update the docs to reflect this change. You'll also need to update the cli help text in the bin file

itszero:
I've updated the changelog now. I might've missed something but it looks like that this object is only used internally and this actually makes the original CLI flag (which is `-c, --config`) works so there's no update in docs/CLI needed.

tannerlinsley:
Right you are! Good catch :)

PR #1131, opened 2019-04-16
Update README.md for packages/react-static-plugin-source-filesystem

marqpdx:
Per notes from the issue log.

## Changes/Tasks

- [ ] Changed code

## Types of changes

## Checklist:


tannerlinsley:
Instead of the extra comment I think just putting the import in the example code would be better.

PR #1135, opened 2019-04-17
Remove webpack-dev-server/client from entry

### LINKED ISSUE #1109: [Bug] Local development server broken
I'm [noticing](https://spectrum.chat/react-static/general/broken-dev-builds-react-static-v6-3-x-and-7-0-x~8e1ac78a-d23f-406c-be05-57010f04368c) that fresh installs of React Static displays an empty page in develop mode (`$ react-static start`) and the console outputs `SyntaxError: The URL 'http:/[http//localhost]:300' is invalid` from SockJS. This occurs after running yarn upgrade in a previously working repo in either RS 6.3.x or 7.0.x. Seems that this is related to the 3.3.0 update of webpack-dev-server ([possibly this change?](https://github.com/webpack/webpack-dev-server/pull/1664)).

Adding this to my project's package.json and reinstalling dependencies fixes development builds. This confirms that something is borked in an update of webpack-dev-server.
```
"resolutions": {
  "react-static/webpack-dev-server": "~3.1.10"
}
```
### Environment

```
  System:
    OS: macOS 10.14.4
    CPU: (4) x64 Intel(R) Core(TM) i5-6287U CPU @ 3.10GHz
    Memory: 228.44 MB / 16.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 10.15.3 - /usr/local/bin/node
    Yarn: 1.15.2 - /usr/local/bin/yarn
    npm: 6.4.1 - /usr/local/bin/npm
  Browsers:
    Chrome: 73.0.3683.86
    Firefox: 66.0.2
    Safari: 12.1
  npmPackages:
    react-ga: ^2.5.7 => 2.5.7 
    react-markdown: ^4.0.6 => 4.0.6 
    react-moment: ^0.8.4 => 0.8.4 
    react-static: ^7.0.7 => 7.0.7 
    react-static-plugin-reach-router: ^7.0.0 => 7.0.0 
    react-static-plugin-sitemap: ^7.0.0 => 7.0.0 
    react-static-plugin-source-filesystem: ^7.0.0 => 7.0.0
```

1.  `npx react-static create`
2.  create site and cd in
3.  `yarn upgrade` to install latest dependencies
4.  `yarn start`
5.  navigate to `http://localhost:3000`
6.  See broken site? Open console.

SleeplessByte:
## Description

According to the [webpack-dev-server documentation](https://webpack.js.org/guides/hot-module-replacement/ ), the client should not be an entry. It seems to be an old way of configuring (I found this [article from 2016](https://medium.com/@rajaraodv/webpacks-hmr-react-hot-loader-the-missing-manual-232336dc0d96) that shows the client as an entry point).

When this line is present, the webpack-dev-server is actually required _twice_, as you can see in the following image:

1. once from the entry (`'webpack-dev-server/client?...'`)
2. once from the plugin (`'../../webpack-dev-server/client?...'`)

This would normally not be an issue. Identical requires are only executed once. However, `webpack-dev-server` as a plugin expects you to write:
```javascript
devServer: {
  host: 'hostname-without-scheme',
  port: number
}
```
And the manual `entry` expects you to write:
```javascript
devServer: {
  host: 'http://hostname-without-scheme',
  port: number
}
```
This results in two unique requires.

This might be a change they made recently, this might be a cross-env issue. I'm not certain. When this line is removed from the configuration, the HMR plugin correct _injects_ this entry into the configuration, only once.

Fixes #1109 

## Changes/Tasks

- [x] Changed code

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [X] I have updated the documentation accordingly

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - `react-static`: Alternative dev server port is now accurate
 - `react-static`: Dev server location is now logged in blue for more visibility
+- `react-static`: Dev server is only required once
 
 # 7.0.8
 
--- a/packages/react-static/package.json
+++ b/packages/react-static/package.json
@@ -98,7 +98,7 @@
     "url-loader": "^1.1.2",
     "webpack": "^4.25.1",
     "webpack-bundle-analyzer": "^3.0.3",
-    "webpack-dev-server": "^3.1.10",
+    "webpack-dev-server": "^3.3.1",
     "webpack-flush-chunks": "^2.0.1",
     "webpack-node-externals": "^1.7.2"
   },
--- a/packages/react-static/src/static/webpack/webpack.config.dev.js
+++ b/packages/react-static/src/static/webpack/webpack.config.dev.js
@@ -22,9 +22,6 @@ export default function({ config }) {
     },
     context: path.resolve(__dirname, '../../../node_modules'),
     entry: [
-      `webpack-dev-server/client?${config.devServer.host}:${
-        config.devServer.port
-      }`,
       'webpack/hot/only-dev-server',
       ...(config.disableRuntime
         ? []
--- a/yarn.lock
+++ b/yarn.lock
@@ -3711,7 +3711,7 @@ compression@1.7.3:
     safe-buffer "5.1.2"
     vary "~1.1.2"
 
-compression@^1.5.2:
+compression@^1.7.4:
   version "1.7.4"
   resolved "https://registry.yarnpkg.com/compression/-/compression-1.7.4.tgz#95523eff170ca57c29a0ca41e6fe131f41e5bb8f"
   integrity sha512-jaSIDzP9pZVS4ZfQ+TzvtiWhdpFhE2RDHz8QJkpX9SIpLq88VueF5jJw6t+6CUQcAoA6t+x89MLrWAqpfDE8iQ==
@@ -3747,7 +3747,7 @@ config-chain@^1.1.11:
     ini "^1.3.4"
     proto-list "~1.2.1"
 
-connect-history-api-fallback@^1.3.0:
+connect-history-api-fallback@^1.6.0:
   version "1.6.0"
   resolved "https://registry.yarnpkg.com/connect-history-api-fallback/-/connect-history-api-fallback-1.6.0.tgz#8b32089359308d111115d81cad3fceab888f97bc"
   integrity sha512-e54B99q/OUoH64zYYRf3HBP5z24G38h5D3qXu23JGRoigpX5Ss4r9ZnDk3g0Z8uQC2x2lPaJ+UlWBc1ZWBWdLg==
@@ -4381,13 +4381,6 @@ decamelize@^1.1.0, decamelize@^1.1.1, decamelize@^1.1.2, decamelize@^1.2.0:
   resolved "https://registry.yarnpkg.com/decamelize/-/decamelize-1.2.0.tgz#f6534d15148269b20352e7bee26f501f9a191290"
   integrity sha1-9lNNFRSCabIDUue+4m9QH5oZEpA=
 
-decamelize@^2.0.0:
-  version "2.0.0"
-  resolved "https://registry.yarnpkg.com/decamelize/-/decamelize-2.0.0.tgz#656d7bbc8094c4c788ea53c5840908c9c7d063c7"
-  integrity sha512-Ikpp5scV3MSYxY39ymh45ZLEecsTdv/Xj2CaQfI8RLMuwi7XvjX9H/fhraiSuU+C5w5NTDu4ZU72xNiZnurBPg==
-  dependencies:
-    xregexp "4.0.0"
-
 decode-uri-component@^0.2.0:
   version "0.2.0"
   resolved "https://registry.yarnpkg.com/decode-uri-component/-/decode-uri-component-0.2.0.tgz#eb3913333458775cb84cd1a1fae062106bb87545"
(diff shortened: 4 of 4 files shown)


tannerlinsley:
It appears that even after this change, I am seeing 2 versions of `only-dev-server` and `dev-server`. Not sure where they're coming from yet, but man... I hate hot-reloading.

SleeplessByte:
@tannerlinsley that's correct, that's a different "issue". It's probably because you explicitely added `'webpack/hot/only-dev-server'`, which I don't know about yet :)

I'll read about that next if you'd like and submit another PR.

tannerlinsley:
Ah. I added it because it's mean to override the standard dev server code and not reload the page when things error. It's primarily for supporting things like react-hot-loader that offer inline-reloading of components.

SleeplessByte:
Got it! I think it's fine (for now) to have both `only-dev-server` and `dev-server`, but IIRC, with webpack four, `dev-server` _should_ be able to support all that.

dgavara:
@tannerlinsley When are you releasing the new version of react-static?

tannerlinsley:
Still waiting on some changes to play out on master with hot-reloading. Soon.

SleeplessByte:
Hey @tannerlinsley. Finally found something sane in the [documentation](https://webpack.js.org/configuration/dev-server/#devserverhot). Combine that with [this](https://webpack.js.org/guides/hot-module-replacement/#enabling-hmr).

In other words: 
- I think you can remove the `hot/only-dev-server` line, and rely on `hotOnly: true` to provide it. Leave the plugin in.
- This means `hot: true` is **incorrect**.

tannerlinsley:
So remove `hot: true` and add `hotOnly: true`?

SleeplessByte:
Not at my regular machine, so have not tested.

Remove these:

- https://github.com/nozzle/react-static/blob/master/packages/react-static/src/static/webpack/webpack.config.dev.js#L26

- https://github.com/nozzle/react-static/blob/master/packages/react-static/src/static/webpack/runDevServer.js#L84

- Add `hotOnly: true`.

Leave everything else the same.

Can you try that out and see if it works as expected? ❤️

tannerlinsley:
I'm on it.

SleeplessByte:
FYI, I'm trying to match [this official configuration](https://webpack.js.org/guides/hot-module-replacement/#enabling-hmr) , but then using the `hotOnly` option 💯

tannerlinsley:
Just tried it an hot-reloading fails silently during runtime. Nothing is triggered.

tannerlinsley:
This is what I have so far:

- `hot: true`
	- injects the hot dev client into the browser
	- shows a single listener
    - hot-reloading events are not receieved in the browser
    - hot reloading doesn't work
- `hotOnly: true`
	- same as previous
- `hot:true, hotOnly: true`
	- same as previous
- Adding `webpack/hot/only-dev-server` to any of the above configurations:
	- injects the hot dev client into the browser
	- injects the hotOnly dev client into the browser
    - show 2 listeners
    - hot-reloading events are received in the browser
    - hot reloading works

tannerlinsley:
I think I know what is happening here:

- `hotOnly` is working correctly in webpack dev server
Screen Shot 2019-04-24 at 3 48 08 PM
- Unfortunately, it's using `require.resolve`. Because of this, there end up being 2 different versions/instances of the client, so when the hot events come in to the browser, one instance is receiving them, but has no subscriptions, and the other instance that has the subscription doesn't receive the events. 

If webpack-dev-server's code doesn't use `require.resolve` it works perfectly. Clearly that's not an option (I doubt they would accept a PR like that), so from the other end, I want to figure out which module is trying to import stuff from the webpack-dev-server and instantiating it again. Maybe react-hot-loader?

SleeplessByte:
Ah, so https://github.com/nozzle/react-static/pull/1135#issuecomment-486437870 means that the first two options _do_ work correctly if it weren't using `require.resolve`?

> one instance is receiving them, but has no subscriptions, and the other instance that has the subscription doesn't receive the events.

This makes sense that this would be the issue -- 

Why doesn't `require.resolve` work? As in, can't we just use that as well?

SleeplessByte:
`react-hot-loader` should probably be used _on top_ of hotOnly, right?

tannerlinsley:
Yes. I don't even know if it's RHL's fault. Something is resolving to a separate copy of the dev server code and not syncing up on the socket instance.

tannerlinsley:
Here, you can see webpack emitting a hot-reload event through the emitter module:
Screen Shot 2019-04-25 at 10 24 28 AM
And here, you can see where that event is meant to be received and trigger a `check` (which is what would eventually lead to `hot.module.check/accept` calls running: Screen Shot 2019-04-25 at 10 24 50 AM

Unfortunately, the underlying emitter is not the same instance between the two, so they never communicate.

tannerlinsley:
This could even be due to the dev environment of react-static. When npm-linked, webpack (which is running from within react-static files) will resolve to dependencies within react-static instead of the linked project. It's very possible that this isn't even a problem in actual projects that aren't npm-linked, but I don't want that to be an excuse for this not working. We should be able to develop in react-static and on linked projects seamlessly.

Maybe an alias to resolve any emitter imports to the same file would be a good idea?

SleeplessByte:
> This could even be due to the dev environment of react-static. When npm-linked, webpack (which is running from within react-static files) will resolve to dependencies within react-static instead of the linked project. It's very possible that this isn't even a problem in actual projects that aren't npm-linked, but I don't want that to be an excuse for this not working. We should be able to develop in react-static and on linked projects seamlessly.

I wholeheartedly agree with you! This should work regardless of the environment. 

> Maybe an alias to resolve any emitter imports to the same file would be a good idea?

If that solves this for now, then sure! I'd rather have less duplication / 2 servers running.

SleeplessByte:
https://github.com/nozzle/react-static/pull/1135#issuecomment-486743734

Is this in the same directory? because then it's indeed not the same path.

tannerlinsley:
Figured it out. An alias in dev mode to resolve `webpack/hot/emitter` to the same module does the trick!

PR #1165, opened 2019-05-01
use latest versions for typescript template

sw-yx:
## Description

`react-static create` somehow doesnt install latest patch versions of RS and this is bad for initial experience

## Changes/Tasks

- [x] Changed code

## Motivation and Context

an attempt to fix https://github.com/nozzle/react-static/issues/1163 for good - what do we think?

## Types of changes

## Checklist:


SleeplessByte:
`latest` breaks completely when removing package.lock / yarn.lock or force upgrade.

PR #1172, opened 2019-05-05
less plugin pointing to incorrect main file

arlair:
`src` has a `node.api.js` file, not a `index.js` file, so `main` should be `node.api.js`.

The 7.0.10 npm install doesn't seem to have a compiled `node.api.js`, as there is only a copy in `src` and none in the root. This is probably the cause of this issue: https://github.com/nozzle/react-static/issues/1150

## Changes/Tasks

- [x] Update main in packages.json.

## Motivation and Context

Fix npm package.

https://github.com/nozzle/react-static/issues/1150

## Types of changes

## Checklist:

### DIFF
--- a/packages/react-static-plugin-less/package.json
+++ b/packages/react-static-plugin-less/package.json
@@ -1,7 +1,7 @@
 {
   "name": "react-static-plugin-less",
   "version": "7.0.10",
-  "main": "index.js",
+  "main": "node.api.js",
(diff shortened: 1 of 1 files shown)


kigorw:
hi, you have the same issue with `react-static-plugin-stylus` please fix there as well

SleeplessByte:
We accept PRs!

PR #1193, opened 2019-05-19
fix: HMR of real CSS files in development

### LINKED ISSUE #1169: [Bug] CSS Hot Reload Not Working
### Environment

  **System**:
    OS: Windows 10
    CPU: (8) x64 Intel(R) Core(TM) i7-6700HQ CPU @ 2.60GHz
    Memory: 2.11 GB / 7.87 GB
  **Binaries**:
    Node: 10.15.3 - C:\Program Files\nodejs\node.EXE
    npm: 6.4.1 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: 44.17763.1.0
    Internet Explorer: 11.0.17763.1
  **NPM**:
    "react": "^16.8.2",
    "react-dom": "^16.8.2",
    "react-static": "^7.0.0",
    "react-static-plugin-reach-router": "^7.0.0",
    "react-static-plugin-sitemap": "^7.0.0",
    "react-static-plugin-source-filesystem": "^7.0.0"

1. npm i -g react-static
2. react-static create
3. choose basic example
4. change app.css CSS to trigger reload

### Expected Behavior
Reloaded CSS displayed on screen. 

JS changes are reflected correctly but CSS does not apply on screen until I manually refresh. After experiencing this issue I created a fresh basic template and this is still not working there. 

After editing app.css the console shows - File changed: \src\app.css
Updating bundle...
[✓] Bundle Updated (0.7s)

so changes are being detected just not displayed.

ScriptedAlchemy:
Its been a long and hard road. But HMR should now work reliably for any and all types of style content. Allowing users to reload real CSS files. No more style-loader needed

## Description

Re-adding extract-css-chunks HMR system back to react static development mode after joining forces with mini-css authors

## Changes/Tasks

- [x] Changed code

## Motivation and Context

I get a staggering amount of emails from GitHub issues on react static. Id like less on the CSS subject

## Types of changes

## Checklist:

fixes #1169

### DIFF
--- a/packages/react-static-plugin-css-modules/package.json
+++ b/packages/react-static-plugin-css-modules/package.json
@@ -21,7 +21,7 @@
   },
   "dependencies": {
     "css-loader": "^2.0.1",
-    "extract-css-chunks-webpack-plugin": "^4.3.2"
+    "extract-css-chunks-webpack-plugin": "^4.5.1"
   },
   "gitHead": "ca9f2162257fe67a8b4ffcbb1350695345940da4"
 }
--- a/packages/react-static-plugin-css-modules/src/node.api.js
+++ b/packages/react-static-plugin-css-modules/src/node.api.js
@@ -4,7 +4,6 @@ export default (options = {}) => ({
   webpack: (config, { stage }) => {
     let loaders = []
 
-    const styleLoader = { loader: 'style-loader' }
     const cssLoader = {
       loader: 'css-loader',
       options: {
@@ -16,7 +15,16 @@ export default (options = {}) => ({
 
     if (stage === 'dev') {
       // Dev
-      loaders = [styleLoader, cssLoader]
+      loaders = [
+        {
+          loader: ExtractCssChunks.loader,
+          options: {
+            hot: true,
+            reloadAll: true,
+          },
+        },
+        cssLoader,
+      ]
     } else if (stage === 'node') {
       loaders = [
         {
--- a/packages/react-static-plugin-less/package.json
+++ b/packages/react-static-plugin-less/package.json
@@ -18,7 +18,7 @@
   "dependencies": {
     "autoprefixer": "^9.5.1",
     "css-loader": "^2.1.0",
-    "extract-css-chunks-webpack-plugin": "^4.3.2",
+    "extract-css-chunks-webpack-plugin": "^4.5.1",
     "less": "^3.9.0",
     "less-loader": "^4.1.0",
     "postcss-flexbugs-fixes": "^4.1.0",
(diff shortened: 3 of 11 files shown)


ScriptedAlchemy:
@tannerlinsley any ideas why Travis is whining?

SleeplessByte:
@ScriptedAlchemy Travis has been whining for a while, make sure the tests run locally and you're fine.

ScriptedAlchemy:
Thanks. 
She’s all good in the hood when running locally

ScriptedAlchemy:
Yeah i ran tests and formatting as soon as it failed assuming I'd broken something. All clear locally.

Ill try to dig into the issue when time frees up

silvenon:
Just out of curiosity, what does `reloadAll: true` do _in combination with_ `hot: true`? I see that that is the case with CSS Modules plugin.

SleeplessByte:
```javascript
{
  hot: true, // if you want HMR - we try to automatically inject hot reloading but if it's not working, add it to the config
  reloadAll: true, // when desperation kicks in - this is a brute force HMR flag
}
```

https://github.com/faceyspacey/extract-css-chunks-webpack-plugin

My guess is that it's copied from the configuration. My _hope_ is that it uses `hot` and only `reloadAll` if it fails. This might break the error overlay for css errors tho.

silvenon:
Yeah, that's my guess as well based on the docs, but it would be great to test this just to ensure we don't lose HMR. While testing locally my own setup it seems that adding `reloadAll` makes HMR somewhat slower. 😕 So unless there's a specific reason why CSS Modules have `reloadAll` and regular CSS doesn't, maybe we don't need it? But @ScriptedAlchemy probably knows about this more than I do.

ScriptedAlchemy:
Reload all is a simple one. Literally loads all css files again. Not recompile just reload them in browser. Reason is, CSS modules are hard. 

However, if you are using css-loader, then I’ve worked together with webpack to detect if you are using css modules. 

I also maintain mini-css, in theory we don’t seem to have issues with reloadAll and is more on the rare side, however builds are very hard to guess what’s going to work. 

If a large number of the user population can confirm that reloadAll: false still works with both css and css modules. Then we can flip it. 

Again reload all is very primitive, appends a timestamp to every stylesheet found in the DOM and just loads the resource again. 

Excuse the primitive nature

SleeplessByte:
I think the main question was: **Why** both `hot` and `reloadAll` 🗡

ScriptedAlchemy:
Hot enables the loader, reloadAll tweaks the loafers reloading strategy. If hot is false then no code is ever appended. Like in production builds

SleeplessByte:
Perfect, that should answer @silvenon question :)

silvenon:
I just wanted to add that I've been using only `hot: true` in my project with CSS Modules for a while now and it seems to work fine, but I understand that enabling `reloadAll` is safer. 👍

ScriptedAlchemy:
Very thrilled to hear that, seems like a simple task but getting css modules to “just work” was an undertaking that required a huge rewrite. Everyone who reports HMR with hot only makes me smile.

silvenon:
Thanks for all your great work on extract-css-chunks-webpack-plugin! 🥇

PR #1213, opened 2019-06-04
Add react-static-plugin-stylus

JonDum:
## Description

Adds a plugin for loading stylus `.styl` files.

## Changes

- [x] Adds stylus plugin

## Motivation and Context

Because I prefer Stylus and there wasn't one yet.

## Types of changes

## Checklist:


JonDum:
This is basically copy pasted from the less/sass loaders with two distinctions.

1) I added a `cssLoaderOptions` that spreads over to the `css-loader` options. This might even be a good standard to add to the other css plugins as it allows more flexibility to users.

2) I use ExtractCssChunks in both dev and prod. On my team we've had a few nasty surprises going from dev to prod when we used to have `style-loader` only for dev. Turns out that you can have some odd css selector specificity issues going from `<style>` tags to `<link>` style sheets. We found it's better to just use ExtractCssChunks in both dev and prod to make the environments as similar as possible.

tannerlinsley:
Just make sure that the react-static-plugin-stylus package name is available, or that `tannerlinsley` is added as an admin on it and this will get published with the next release.

PR #1218, opened 2019-06-10
[Reach Router] Add base path to server location URL

nddery:
Fixes https://github.com/nozzle/react-static/issues/1182 - where static export doesn't work when you set a `basePath` on your config.

## Description
Reach router (plugin) specifically makes the location URL absolute - meaning that given a `staticInfo.path` of `blog/post/79`, it'll transform it to `/blog/post/79` - which is outside of our app, hence rendering nothing. The `ServerLocation` lives outside of the main router, hence it needs to provide the full URL, with the base path.

## Changes/Tasks
- [x] Add `basepath` to `ServerLocation` URL.

## Types of changes

## Checklist:
I haven't added any test, as I didn't see any in the Reach Router plugin. For the CHANGELOG summary, should I just add a section to it (there's no "unreleased" section) ?


sw-yx:

Issue #1220, opened 2019-06-10
[Bug] v7.1.0 Browserlist warning in dev environment


kirklikecaptain:
After updating to react-static v7.1.0, the warning below is displayed in the terminal when the dev server is starting. It doesn't appear to cause any issues so far, but it's worth mentioning.

```
[✓] Templates Built
Bundling Application...
Fetching Site Data...
[✓] Site Data Downloaded (0.1s)

  Replace Autoprefixer browsers option to Browserslist config.
  Use browserslist key in package.json or .browserslistrc file.

  Using browsers option cause some error. Browserslist config
  can be used for Babel, Autoprefixer, postcss-normalize and other tools.

  If you really need to use option, rename it to overrideBrowserslist.

  Learn more at:
  https://github.com/browserslist/browserslist#readme
  https://twitter.com/browserslist



  Replace Autoprefixer browsers option to Browserslist config.
  Use browserslist key in package.json or .browserslistrc file.

  Using browsers option cause some error. Browserslist config
  can be used for Babel, Autoprefixer, postcss-normalize and other tools.

  If you really need to use option, rename it to overrideBrowserslist.

  Learn more at:
  https://github.com/browserslist/browserslist#readme
  https://twitter.com/browserslist



  Replace Autoprefixer browsers option to Browserslist config.
  Use browserslist key in package.json or .browserslistrc file.

  Using browsers option cause some error. Browserslist config
  can be used for Babel, Autoprefixer, postcss-normalize and other tools.

  If you really need to use option, rename it to overrideBrowserslist.

  Learn more at:
  https://github.com/browserslist/browserslist#readme
  https://twitter.com/browserslist


[✓] Application Bundled (20.5s)
```

niboc:
How is it fixed?

theWebalyst:
I just installed RS v7.1.0 and if I make a new 'basic' project, sometimes the dev server works, but then I run it again and it fails. Sometimes it fails right away and I don't see why. 

When it fails I get the warning as described in the OP, but in my case the website is broken. I see the page display for a moment and then a black screen with the following:

```
An internal error occured!

RoutesInner@http://localhost:3000/main.js:53224:13
renderWithHooks@http://localhost:3000/main.js:38775:22
mountIndeterminateComponent@http://localhost:3000/main.js:40931:17
beginWork@http://localhost:3000/main.js:41566:20
performUnitOfWork@http://localhost:3000/main.js:45610:16
[...]

PR #1226, opened 2019-06-13
Use 'overrideBrowserslist' option instead of 'browsers' in autoprefixer

jxom:
## Description

As of [autoprefixer@9.6.0](https://github.com/postcss/autoprefixer/releases/tag/9.6.0), autoprefixer recommends to use a `.browserslistrc` file or a `browserslist` attribute in package.json. Currently, if you try to run or build a react-static project, it will [warn you and tell you to do the above (#1220)](https://github.com/nozzle/react-static/issues/1220).

If we remove the `browsers` option in the autoprefixer, this will introduce a breaking change as consumers will need to create a browserslist config. For now, I have renamed the `browsers` option to `overrideBrowserslist` to avoid any breaking changes, and in the next major React Static release, maybe we remove this option in favour of the browserslist config file. Keen to hear your thoughts!

## Changes/Tasks

- [x] Rename `browsers` option to `overrideBrowserslist` in the autoprefixer config.
- [x] Update autoprefixer to ^9.6.0

## Motivation and Context

Solves #1220.

## Types of changes

## Checklist:

### DIFF
--- a/packages/react-static-plugin-less/package.json
+++ b/packages/react-static-plugin-less/package.json
@@ -16,7 +16,7 @@
     "@babel/preset-env": "^7.4.3"
   },
   "dependencies": {
-    "autoprefixer": "^9.5.1",
+    "autoprefixer": "^9.6.0",
     "css-loader": "^2.1.0",
     "extract-css-chunks-webpack-plugin": "^4.5.1",
     "less": "^3.9.0",
--- a/packages/react-static-plugin-less/package.json
+++ b/packages/react-static-plugin-less/package.json
@@ -24,5 +24,11 @@
     "postcss-flexbugs-fixes": "^4.1.0",
     "semver": "^6.0.0"
   },
+  "browserslist": [
+    ">1%",
+    "last 4 versions",
+    "Firefox ESR",
+    "not ie < 9"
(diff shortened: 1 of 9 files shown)


digitalkaoz:
im assuming they have a hierarchy like babel? Then i would first try to include the config file from the users workspace, otherwise we load the default one from react-static?

cortopy:
`overrideBrowserslist` feels very cumbersome to me (as a newcomer). I want to use my own browserlist and with this change, I would have to reconfigure every webpack plugin that uses the list in order to "override the override"

sw-yx:
anyone one to send in an alternative PR?

jxom:
Fair call. I've committed an update which adds browserslist config to react-static to fallback on instead - allowing the user to specify their own browserslist config.

digitalkaoz:
@jxcom im not sure your latest change is correct. I cant see the fallback scenario

sw-yx:

SleeplessByte:
Closed in favour of #1283 

Fixed via https://github.com/react-static/react-static/pull/1283/commits/7c244e45dbdbf8dd12cc4c09ad412253db869cf2

PR #1232, opened 2019-06-17
Use contenthash over hash for deterministic hashes

nddery:
## Description
According to Webpack documentation, `hash` is unique per build. When building on multiple machines, this hash is different on all builds. Using `contenthash` ensure we get deterministic hashes between multiple machines.
https://webpack.js.org/configuration/output/#outputfilename

Webpack also recommend using `contenthash` for long term caching.
https://webpack.js.org/guides/caching/#output-filenames

This is fixable in "user-land" by updating their Webpack config, we've done so like below for now, but I think this would be nice to work out-of-the-box.

```js
// node.api.js
export default () => ({
  webpack: (config, { defaultloaders }) => {
    if (config.output.filename !== 'static-app.js') {
      config.output.filename = '[name].[contenthash:8].js'
    }

    return config
  }
}) 
```

## Changes/Tasks
- [x] Update `config.output.filename` to use `contenthash` over `hash`

## Motivation and Context
We are currently transitioning from a monolith Rails application to a static site and during this transition, we are building a React Static site at deploy time. Each server in our load balancer builds the same React Static application and should be able to serve it. When a request is made for a page, the markup could be served from one server while assets get's served from another. With non-deterministic hashes, we end up with a bunch of 404's.

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - `react-static-plugin-stylus` A plugin for using stylus
 - Added a styled-components guide
 - Added `react-static-plugin-google-tag-manager` to the list of 3rd party plugins
+- `react-static`: Use `contenthash` over `hash` in output filename for deterministic filenames
 
 ### Improved
 
--- a/packages/react-static/src/static/webpack/webpack.config.prod.js
+++ b/packages/react-static/src/static/webpack/webpack.config.prod.js
@@ -77,7 +77,7 @@ function common(state) {
           require.resolve('../../bootstrapApp'),
         ],
     output: {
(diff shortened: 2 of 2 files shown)


nddery:
One thing I didn't mention in the description, I haven't updated `chunkHash` to `contentHash` - they seem to work but maybe it would be best to also switch over ? Seems like Webpack [recommend doing so](https://github.com/webpack/webpack.js.org/issues/2096) ? I can update this PR if this makes sense to you!

nddery:
Well, seems like this isn't working in the end - I had tested this and was convinced it was... Oh well. I'll close this pull requests for now, and will open an issue instead for now.

Issue #1177, opened 2019-05-07
FOUC using the SASS plugin


virtualstyle:
I'm not using styled components. I'm using the SASS plugin. Some minor tweaks to the webpack config. On dev (yarn start), the loading component shows nicely, all good. On build, staging or prod, never see a loading indicator, get unstyled content that rapidly moves around as my styles are applied.

Google searches aren't turning up much, there's the router.subscribe method, which seems to be very old, then something about a Loading hoc, but nothing I can get working. About to give up on the package... what's up with simple loading handling on build rather than dev server?

sw-yx:
hmm. check the generated html is loading correctly. the sass plugin is really simple - if youre familiar with the webpack loader: https://github.com/nozzle/react-static/blob/57ecc92a63182aab91112be5c3e787c338308b09/packages/react-static-plugin-sass/src/node.api.js#L23-L38 PR's welcome to fix though but i think @tannerlinsley uses this actively

virtualstyle:
Well, everything loads correctly eventually, because the site/app is working. The problem is an FOUC, which doesn't happen on dev. So the sass plugin must indeed be getting me my compiled CSS. But I'm NOT getting to see a loading indicator while it does so, I get unstyled content, which jumps around all crazy while css loads and is applied. Never see a flash of my loading component. But all the code loads fine, site works fine. The problem is FOUC.

sw-yx:
sorry, i got nothing 😐 i use the styled components one myself.

silvenon:
@virtualstyle do you have a repo I could try?

virtualstyle:
No repo, new project. I'm going with https://github.com/markdalgleish/static-site-generator-webpack-plugin, I like getting my webpack.config back in my hands.

Reggino:
I'm running into the same issue. I think this should be re-opened

sw-yx:
@Reggino have a repro?

Reggino:
I'll try to setup a demo in a stand-alone repo.

But it seems to be reproducible by disabling Javascript in the browser and then opening a built page: the '.scss' file loaded in the used modules don't seem to embedded in the intially loaded css file.

Reggino:
Reproduction is pretty straightforward:

- Scaffold a fresh app
- Add `sass-plugin`
- Load sass file in page
- Build it.

This setup can be found (here)[https://github.com/Reggino/react-static-issue-1177]

[...]

PR #1196, opened 2019-05-20
Fix publicPath handling

gbrsales:
`publicPath` should be read from `REACT_STATIC_PUBLIC_PATH` and not from `REACT_STATIC_ASSETS_PATH`.

Fixes #1141.

PR #1249, opened 2019-07-02
Prevent FOUC in sass plugin

Reggino:
When using the Sass plugin, a Flash Of Unstyled Content may appear in a built application.

This seems be caused by the missing configuration of Webpack's [Split Chunk Optimization](https://webpack.js.org/plugins/split-chunks-plugin) when using this plugin. 

The default CSS loader loads all CSS in a single chunk by default (when the `extractCssChunks` configuration option isn't set) [See this related source code](https://github.com/nozzle/react-static/blob/master/packages/react-static/src/static/webpack/webpack.config.prod.js#L55), but the SASS module is missing this configuration / is not not patching up the related configuration.

This PR should fix that and fixes issue #1177.

### DIFF
--- a/packages/react-static-plugin-sass/src/node.api.js
+++ b/packages/react-static-plugin-sass/src/node.api.js
@@ -75,6 +75,10 @@ export default ({ includePaths = [], ...rest }) => ({
       use: loaders,
     })
 
+    if (config.optimization.splitChunks.cacheGroups.styles) {
+      config.optimization.splitChunks.cacheGroups.styles.test = /\.(c|sc|sa)ss$/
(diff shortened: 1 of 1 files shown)


sw-yx:
merging you in but i dont control the releases fyi. cheers for the contrib!

Reggino:
@SleeplessByte Could you create a new release for this plugin? Thanks!

SleeplessByte:
Waiting on https://github.com/react-static/react-static/pull/1253#issuecomment-514019300 / #1196, as master is currently "broken" for some configurations!

PR #1219, opened 2019-06-10
Node 12 (stable) support

nddery:
Update `node-sass` to `^4.12.0` and instruct yarn to install `fsevents@^1.2.9`, which allows a fresh install on Node 12.

## Description
Per https://github.com/sass/node-sass/issues/2632#issuecomment-487247910 `node-sass@^4.12.0` supports Node 12. `node-sass` is a direct dependency of `react-static-plugin-sass`.

Per https://github.com/fsevents/fsevents/issues/272#issuecomment-487645485 `fsevents@^1.2.9` supports Node 12. `fsevents` is a transitive dependency of a couple of direct and indirect dependencies.

```
❯ yarn why fsevents
yarn why v1.15.2
[1/4] 🤔  Why do we have the module "fsevents"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "fsevents@1.27"
info Reasons this module exists
   - "_project_#chokidar" depends on it
   - Hoisted from "_project_#chokidar#fsevents"
   - Hoisted from "_project_#react-static#onchange#chokidar#fsevents"
   - Hoisted from "_project_#react-static#webpack#watchpack#chokidar#fsevents"
   - Hoisted from "_project_#react-static#babel-jest#@jest#transform#jest-haste-map#fsevents
```

## Changes/Tasks
- [x] Upgraded to `node-sass@4.12.0`
- [x] Upgraded to `fsevents@1.2.9`

## Motivation and Context
I wanted Travis to run properly.

## Types of changes
- [ ] Refactoring/add tests (refactoring or adding test which isn't a fix or add a feature)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:
- [ ] I have updated the documentation accordingly
- [ ] I have updated the CHANGELOG with a summary of my changes
- [ ] My changes have tests around them

PR #1250, opened 2019-07-03
Don't use a relative path to resolve Default404 template

jviotti:
The current approach assumes that the current directory when running
`react-static` is the root path, which might not be the case if you
customize the root.

This came up as I'm developing a CLI module that embeds `react-static`,
takes some command line arguments, spawns `react-static` using files
inside the CLI module, and stores the results in the user's destination
directory.

When the tool (and therefore `react-static`) are installed globally, my
`ROOT` is:

```
/Users/jviotti/.nvm/versions/node/v9.11.2/lib/node_modules/<my-cli>
```

And `__dirname` from `src/static/getRoutes.js` becomes:

```
/Users/jviotti/.nvm/versions/node/v9.11.2/lib/node_modules/<my-cli>/node_modules/react-static/lib/static
```

Given those two variables, the code in `getRoutes.js` will evaluate the
`Default404` template path to:

```
node_modules/react-static/lib/browser/components/Default404
```

Which is a relative path, therefore `react-static` will try to find
`Default404` in a local installation of `react-static` in the place the
user ran the CLI, which is not expected.

Making the path absolute ensures that `react-static` always defaults the
404 template to the one it itself ships, instead of accidentally cross
referencing another version of itself.

### DIFF
--- a/packages/react-static/src/static/getRoutes.js
+++ b/packages/react-static/src/static/getRoutes.js
@@ -46,12 +46,9 @@ export default async function getRoutes(state, callback = d => d) {
     if (!has404 && !incremental) {
       allNormalizedRoutes.unshift({
         path: '404',
-        template: path.relative(
-          state.config.paths.ROOT,
-          path.resolve(
-            __dirname,
-            path.join('..', 'browser', 'components', 'Default404')
-          )
+        template: path.resolve(
+          __dirname,
+          path.join('..', 'browser', 'components', 'Default404')
         ),
       })
     }
--- a/packages/react-static/src/static/components/BodyWithMeta.js
+++ b/packages/react-static/src/static/components/BodyWithMeta.js
@@ -1,6 +1,6 @@
 import React from 'react'
-import { pathJoin, makePathAbsolute } from '../../utils'
 import jsesc from 'jsesc';
+import { pathJoin, makePathAbsolute } from '../../utils'
 
 const generateRouteInformation = embeddedRouteInfo => ({
   __html: `
--- a/packages/react-static/src/static/components/__tests__/__snapshots__/BodyWithMeta.test.js.snap
+++ b/packages/react-static/src/static/components/__tests__/__snapshots__/BodyWithMeta.test.js.snap
@@ -13,7 +13,7 @@ exports[`BodyWithMeta when route is a redirect route 1`] = `
     </div>
   </body>
 </Component>
-`
+`;
 
 exports[`BodyWithMeta when route is a static route 1`] = `
 <Component
@@ -30,7 +30,7 @@ exports[`BodyWithMeta when route is a static route 1`] = `
       dangerouslySetInnerHTML={
         Object {
           "__html": "
-    window.__routeInfo = {\\"routeDate\\":\\"here\\"};",
+    window.__routeInfo = JSON.parse(\\"{\\\\\\"routeDate\\\\\\":\\\\\\"here\\\\\\"}\\");",
         }
       }
       type="text/javascript"
@@ -49,4 +49,4 @@ exports[`BodyWithMeta when route is a static route 1`] = `
     />
   </body>
 </Component>
-`
+`;
--- a/packages/react-static/src/static/extractTemplates.js
+++ b/packages/react-static/src/static/extractTemplates.js
@@ -17,9 +17,7 @@ export default (async function extractTemplates(state) {
       return
     }
 
-    route.template = slash(
-      path.resolve(config.paths.ARTIFACTS, route.template)
-    )
+    route.template = slash(path.resolve(config.paths.ARTIFACTS, route.template))
 
     // Check if the template has already been added
     const index = templates.indexOf(route.template)
--- a/packages/react-static/src/static/getConfig.js
+++ b/packages/react-static/src/static/getConfig.js
@@ -124,15 +124,15 @@ export function buildConfig(state, config = {}) {
   let assetsPath = ''
   if (process.env.REACT_STATIC_ENV === 'development') {
     basePath = cleanSlashes(config.devBasePath)
-    assetsPath = config.devAssetsPath || paths.assets
+    assetsPath = config.devAssetsPath || paths.assets || assetsPath
   } else if (state.staging) {
     siteRoot = cutPathToRoot(config.stagingSiteRoot || '/', '$1')
     basePath = cleanSlashes(config.stagingBasePath)
-    assetsPath = config.stagingAssetsPath || paths.assets
+    assetsPath = config.stagingAssetsPath || paths.assets || assetsPath
   } else {
     siteRoot = cutPathToRoot(config.siteRoot, '$1')
     basePath = cleanSlashes(config.basePath)
-    assetsPath = config.assetsPath || paths.assets
+    assetsPath = config.assetsPath || paths.assets || assetsPath
   }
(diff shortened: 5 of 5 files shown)


sw-yx:

SleeplessByte:

jviotti:

packages/react-static/src/static/getRoutes.js
```diff
@@ -46,12 +46,9 @@ export default async function getRoutes(state, callback = d => d) {
     if (!has404 && !incremental) {
       allNormalizedRoutes.unshift({
         path: '404',
-        template: path.relative(
-          state.config.paths.ROOT,
```
SleeplessByte: Why was this removed?

`resolve` should probably _still_ use this `root` path?

(p.s. should now be running travis again).

jviotti: The result of `path.resolve(__dirname, path.join('..', 'browser', 'components', 'Default404'))` is an absolute path, which always works fine when passed to `template`. 

Adding `path.relative(ROOT, <path>)` will calculate `<path>` relative to `ROOT`, so the path will be the same, just that now it will have a ton of `../../../` to point to the same directory from the point of view of `ROOT`. 

In the end, the result should be the same, as long as your current working directory is the same as `ROOT`. After this change, I believe everything that worked in the past should still work (lets see what Travis thinks), and it will now still work even if `process.cwd() !== ROOT`:

```sh
$ node
> path.relative('/foo/bar/baz', '/foo/bar/baz/qux')
'qux'
> path.relative('/foo/bar/baz', '/test/bar/baz/qux')
'../../../test/bar/baz/qux'
```

jviotti:
@SleeplessByte Not sure if its just me, but I seem to be getting the same errors Travis is reporting here on `master`, which seems to explain why they look unrelated to my changes here. For example:

```
    TypeError: Cannot read property 'endsWith' of undefined

      142 |
      143 |   // add trailing slash only if assetsPath was supplied, but no trailing slash
    > 144 |   if (assetsPath !== '' && !assetsPath.endsWith('/')) {
          |                                        ^
      145 |     assetsPath = `${assetsPath}/`
      146 |   }
      147 |

      at endsWith (src/static/getConfig.js:144:40)
      at Object.<anonymous> (src/static/__tests__/getConfig.test.js:62:21)
```

Can you confirm it happens on master as well?

jviotti:
@SleeplessByte Looks so. Consider this one:

```
    Snapshot name: `BodyWithMeta when route is a static route 1`

    - Snapshot
    + Received

    @@ -10,11 +10,11 @@
          </div>
          <script
            dangerouslySetInnerHTML={
              Object {
                "__html": "
    -     window.__routeInfo = {"routeDate":"here"};",
    +     window.__routeInfo = JSON.parse("{\\"routeDate\\":\\"here\\"}");",
              }
            }
            type="text/javascript"
          />
          <script
```

The test at `packages/react-static/src/static/components/__tests__/__snapshots__/BodyWithMeta.test.js.snap` definitely expects

```
window.__routeInfo = {\\"routeDate\\":\\"here\\"};",
```

While `packages/react-static/src/static/components/BodyWithMeta.js` has:

```
window.__routeInfo = JSON.parse(${jsesc(JSON.stringify(embeddedRouteInfo), {
```

packages/react-static/src/static/getRoutes.js
SleeplessByte: Thanks for the explanation! 

I didn't really mean, why was `path.relative` removed, but why was `state.config.paths.ROOT` removed. I think we used to be able to change the `ROOT` dir. Overall, I don't really care. All the files _should_ live under `state.config.paths.ROOT` anyway, so `.resolve` _should_ always work 🗡

SleeplessByte:
The tests haven't been running for a while (because travis has been broken for over a month). 

> packages/react-static/src/static/components/__tests__/__snapshots__/BodyWithMeta.test.js.snap

This one can be solved by running `yarn test -u` which will update the snapshot. The new output is the expected one 👍 

I think `if (assetsPath !== '' && !assetsPath.endsWith('/')) {` might need to be changed to

```javascript
if (assetsPath && !assetsPath.endsWith('/')) {
```

Can you see if that works? I think you're correct that this isn't YOUR doing 😓 😅

jviotti:
@SleeplessByte Cool, I'll be adding more commits on top of this PR to try to fix the issues. I applied your suggested fix and updated the snapshots.

jviotti:
@SleeplessByte OK, I got it to pass locally \o/ Can you give a look at the 2 commits I added here?

jviotti:
And Travis is green again!

packages/react-static/src/static/getRoutes.js
jviotti: @SleeplessByte 

> I think we used to be able to change the ROOT dir

We still can do so. This part of the code only kicks in when the user didn't define a 404 page, and therefore we use the default one that comes with the module. The ROOT dir shouldn't make a difference here.

packages/react-static/src/static/__mocks__/defaultConfigProduction.mock.js
```diff
@@ -19,6 +18,11 @@ export default {
     STATIC_DATA: './root/dist/staticData',
   },
   terser: {
-    terserOptions: {},
+    terserOptions: {
+      compress: {},
```
SleeplessByte: Can you explain what this fixed/changed? I just don't know 😅

packages/react-static/src/static/__tests__/getConfig.test.js
```diff
@@ -101,7 +101,7 @@ describe('buildConfig', () => {
     test('when no user-supplied assetsPath exists', () => {
       const { config } = buildConfig({}, { assetsPath: undefined })
 
-      expect(config.assetsPath).toBe('')
+      expect(config.assetsPath).toBeUndefined()
```
SleeplessByte: Does this mean that we _no longer_ default to `''`? I like this/your code better btw!

packages/react-static/src/static/__mocks__/defaultConfigProduction.mock.js
jviotti: I hoped you would know :P From what I can tell, the default config includes these nested empty objects, while the obsolete tests were still expecting just `terserOptions: {}`

packages/react-static/src/static/__tests__/getConfig.test.js
jviotti: That seems to be what's happening at the moment, just that the tests were outdated. Let me give it a shot first anyways, just in case I can also fix it to be `''`

jviotti: @SleeplessByte Done! It should be correctly fixed now

packages/react-static/src/static/__mocks__/defaultConfigProduction.mock.js
jviotti: Hm, I think I also solved this with my other fix. Let me check

jviotti: Indeed! Let me cleanup my commits a bit and we should be good to go.

jviotti:
@SleeplessByte Can you take another look now? The diff is much cleaner now :)

SleeplessByte:
Great!! We'll edit those obsolete tests in a future iteration :)

PR #1251, opened 2019-07-03
cssloader: Rename "browsers" to "overrideBrowserslist"

jviotti:
Running `react-static` currently prints this warning:

```
Replace Autoprefixer browsers option to Browserslist config.
Use browserslist key in package.json or .browserslistrc file.

Using browsers option cause some error. Browserslist config
can be used for Babel, Autoprefixer, postcss-normalize and other tools.

If you really need to use option, rename it to overrideBrowserslist.

Learn more at:
https://github.com/browserslist/browserslist#readme
https://twitter.com/browserslist
```

It goes away by renaming `browsers` to `overrideBrowserslist` as
suggested.


amirhouieh:
When does this come to release?

PR #1181, opened 2019-05-12
Upgrade TypeScript definition, (re)add renderProp for Routes

SleeplessByte:
## Description

This PR started because I needed `useRouteData` _outside_ of the `Routes` component tree. The missing piece was exporting the `routePathContext` context so that a context could be provided everywhere.

When making this change I constantly had to go back to the definitions to figure out what was what. So this PR has the following:
- add `routePathContext` to the exports
- change all the typings (fix missing, incorrect or incomplete typings) (fixes #1163)
- add descriptions to all the `public` functions in the typings
- add back `Routes` `renderProp`, but as render prop, so that the plugins don't break!

⚠️ This is a breaking change. Many of the types I've updated or added are _way_ more strict than the old types. This will probably catch bugs; I don't know if they're _too_ strict. ⚠️

My recommendation is to release it as a `feature` release; even though semantically it should be major. TS is not a first-class citizen yet in terms of react-static, so treating that change as a bugfix and the addition of the context as the feature should be okay.

> **Version 7.0 CHANGELOG**
> The child renderer and corresponding getComponentForPath utility previously provided via the Routes component has been deprecated.

📌 this adds back `getComponentForPath` in `Routes`, but not as `children`. The functionality was not compatible with the plugin system (as `Router` plugins may mutate `children`). Instead it's added as a `render` prop. The reason I added this back is because the `Routes` component does not play well when you want to put something in-between the "Outer" and "Inner" component. Yes, it can be solved with a plugin but this is non-trivial to do.

This enables the following code and also fixes #1153.
```tsx
function App() {
  return (
    <Root>
      <Suspense fallback={<em>Loading...</em>}>
        <Router>
          <Routes default render={
            ({ routePath, getComponentForPath }): React.ReactNode => {
              const Comp = getComponentForPath(routePath)
              return (
                <Wrapper>{Comp}</Wrapper>
              )}
            } />
        </Router>
      </Suspense>
    </Root>
  )
}
```

🚧 Note that I have _only_ tested this with `reach-router`. Can  someone please pull this and test with `react-router`?

## Screenshots (if appropriate):

This shows commentary on all the typings:

typings with commentary

And can now be used to not only validate but also auto-complete the config:

typing the configuration

## Types of changes

- [X] Refactoring/add tests (refactoring or adding test which isn't a fix or add a feature)
- [X] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
[...]

PR #1253, opened 2019-07-05
Fix route template path handling

jviotti:
Consider that we store the artifacts on an OS temporary directory (such
as what you would get with `mktemp -d` or `node-tmp`) which we can
assume for simplicity purposes to be `/tmp/react-static`.

In the snippet above, `route.template` is a path such as
`/home/johndoe/site/lib/pages/main.js`. Therefore `path.relative()` will
calculate the path to `/home/johndoe/site/lib/pages/main.js` from
`/tmp/react-static`, so the result will be something like
`../home/johndoe/site/lib/pages/main.js` which makes no sense and will
cause react-static to fail.

What we really intended is `path.resolve()`, so that the user still gets
`/home/johndoe/site/lib/pages/main.js`.

### DIFF
--- a/packages/react-static/src/static/extractTemplates.js
+++ b/packages/react-static/src/static/extractTemplates.js
@@ -16,9 +16,11 @@ export default (async function extractTemplates(state) {
     if (!route.template) {
       return
     }
+
     route.template = slash(
-      path.relative(config.paths.ARTIFACTS, route.template)
+      path.resolve(config.paths.ARTIFACTS, route.template)
     )
+
     // Check if the template has already been added
     const index = templates.indexOf(route.template)
     if (index === -1) {


sw-yx:
i think you have a better sense of what needs to happen here than i do. leaving this convo for @tannerlinsley

sw-yx:

SleeplessByte:

jviotti:

nddery:
This seems to have introduced a bug where `route.template` will point to `/path/to/site/artifacts/src/containers/Post`. Note the **artifacts** directory in there. I assume the path should be `/path/to/site/src/containers/Post`.

This can be replicated by starting a fresh project (`react-static create`) and linking your local `react-static` (which contains this commit) (`cd my-static-site && yarn link react-static`). From there running `yarn build` should error out with the following error:

> Module not found: Error: Can't resolve '/my-static-site/artifacts/src/containers/Post'

Unsure of the fix here, and of all the scenarios that should be supported. For me, `route.template` isn't absolute, it is the exact value as entered in `static.config.js`.

SleeplessByte:
I figured this wouldn't have been working correctly anyway until #1196 was fixed.
@jviotti can you reproduce the above and fix in a subsequent PR?

nddery:
Unsure a fix for #1196 will fix this in all scenario - or at least the code here might be simplified otherwise.

Given that [`path.resolve`](https://nodejs.org/docs/latest/api/path.html#path_path_resolve_paths) will resolve from right to left, until an absolute path is made and that `config.paths.ARTIFACTS` is an absolute path to the `artifacts` folder, the proposed change will result in either resolving to `route.template` if it (`route.template`) is already absolute or to `config.paths.ARTIFACTS + route.remplate` if it isn't.

```bash
$ node
> const path = require('path')

> path.resolve('/my-static-site/artifacts', '/my-static-site/src/containers/Post') # an absolute route, 👍 
'/my-static-site/src/containers/Post'
> path.resolve('/my-static-site/artifacts', 'src/containers/Post') # a relative route 👎 
'/my-static-site/artifacts/src/containers/Post'
```

If the fix for #1196 makes `route.template` always absolute, then could we simplify the change to the following ?

```diff
- route.template = slash(path.resolve(config.paths.ARTIFACTS, route.template))
+ route.template = slash(route.template)
```

SleeplessByte:
> If the fix for #1196 makes route.template always absolute, then could we simplify the change to the following ?

I prefer this @nddery. Less resolving and more absolute paths here is probably better.

The problem is that we internally only use relative paths (then it works) and @jviotti only uses absolute paths (then it works) and your case here is absolutely valid and breaks it!

nddery:
Would it be safe to assume that if `route.template` is relative, it should resolve against `static.conf.js` location, and if it's absolute, it should be used as-is ? Something like:

```js
route.template = slash(path.resolve('/path/to/static.conf.js', route.template))
```

SleeplessByte:
I am a bit weary of saying YES that's SAFE (because we don't have comprehensive tests around these permutations), but I am okay with making that a breaking change. As long as _build_ artifacts can be stored _outside_ of the repo (per OP / @jviotti).

nddery:
Just to clarify, is the idea to support either relative or absolute path in the `static.config.js`, or to support only one ? Currently the documentation for [documentation](https://github.com/react-static/react-static/blob/master/docs/config.md#route) for a `route` object says to use relative paths - if the idea is to switch to absolute paths everywhere, I can open a PR that updates the documentation for the route object as well as all the [templates](https://github.com/react-static/react-static/tree/master/packages/react-static/templates) config.

SleeplessByte:
I prefer checking if it's path or root relative, or absolute and internally expand them to all be absolute, if we can. Does that make sense?

nddery:
That makes sense, thanks for clarifying things for me!

nddery:
@SleeplessByte re https://github.com/react-static/react-static/pull/1272#issuecomment-519334670 (making this a non-breaking change) the solution proposed above (https://github.com/react-static/react-static/pull/1253#issuecomment-514191625) seems valid to me.

Per the current documentation, `route.template` should be relative to `static.config.js`/the root of the project - so by making the change proposed, we'd keep this behavior, as well as support supplying an absolute path. The resulting `route.template` would however always be absolute, ensuring build artifacts can be stored outside of the repo.

SleeplessByte:
That is perfect.

SleeplessByte:
As a final request @nddery , if you have the time, can you apply that to #1196. It seems that the original author is not around and I'd like to get that moving. In the meanwhile I'll finish up on #1181 and then we're golden for a maintenance release :heart:.

Per our discussions I have also put on the todo list to get starters and examples back up, as well as cleaning up these configurations. The best way forward is probably to _always_ have absolute paths inside react-static core, but allow people to use relative paths in the configuration.

nddery:
Sounds good - I can update the `route.template` documentation to indicate we accept either a root relative path or an absolute path if it make sense ? I'll open a new PR as I don't think I can push to #1196. However, I don't think it fixes using a `devBasePath`, I'll comment over there. I'll try to get that done before leaving today.

PR #1254, opened 2019-07-05
Don't use path.relative() in react-static/src/static

jviotti:
I've been lately hitting various issues due to the way react-static
defaults to relative paths, which tend to break as soon as you tweak
react-static directories (artifacts, dist, root, etc) to locations that
are not the ones react-static recommend.

Most of these issues are automatically resolved by switching to
`path.resolve()`, which always results in absolute paths, and avoids
large `../../` chains if the second argument is absolute.

See: https://github.com/nozzle/react-static/pull/1250
See: https://github.com/nozzle/react-static/pull/1253

### DIFF
--- a/packages/react-static/src/static/generateBrowserPlugins.js
+++ b/packages/react-static/src/static/generateBrowserPlugins.js
@@ -24,7 +24,7 @@ export default async state => {
           : -1
         if (pluginIndex === -1 && browserLocation) {
           pluginImports.push(
-            slash(path.relative(config.paths.ARTIFACTS, browserLocation))
+            slash(path.resolve(config.paths.ARTIFACTS, browserLocation))
           )
           pluginIndex = pluginImports.length - 1
         }
@@ -33,7 +33,7 @@ export default async state => {
 
         // IIF to return the final plugin
         return `{
-        location: "${slash(path.relative(config.paths.ARTIFACTS, location))}",
+        location: "${slash(path.resolve(config.paths.ARTIFACTS, location))}",
         plugins: ${recurse(plugins || [])},
         hooks: ${
           browserLocation
--- a/packages/react-static/src/static/webpack/webpack.config.dev.js
+++ b/packages/react-static/src/static/webpack/webpack.config.dev.js
@@ -46,7 +46,7 @@ export default function({ config }) {
         NODE_MODULES,
         SRC,
         DIST,
-        ...[NODE_MODULES, SRC, DIST].map(d => path.relative(__dirname, d)),
+        ...[NODE_MODULES, SRC, DIST].map(d => path.resolve(__dirname, d)),
         'node_modules',
       ],
       extensions: ['.wasm', '.mjs', '.js', '.json', '.jsx'],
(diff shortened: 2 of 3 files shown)


jviotti:
I'm able to confirm the changes work on my react-static project, but I can't get the test suite to run locally (and Travis seems to have issues as well), so I can't speak for that, but everything should work fine unless there are parts of the code that really expect a relative path and where an absolute path would be invalid, even though that seems unlikely.

sw-yx:
excellent. i dont know enough to be able to merge this freely so unfortunately this needs @tannerlinsley  to glance at it

sw-yx:

SleeplessByte:

jviotti:

SleeplessByte:
@jviotti Let me see if I can fix that ;)

SleeplessByte:
https://travis-ci.com/react-static/react-static/builds/119630460 some of the tests are failing, some of then due to https://github.com/react-static/react-static/pull/1253.

Would you be so kind to try to make them pass? This should run _locally_ as well!

jviotti:
@SleeplessByte I resolved the conflicts. See https://github.com/react-static/react-static/pull/1250. I think these tests are failing on master, strangely enough.

> There should be some tests for this at some point; but I won't enforce that now.

I think the best way to test this is to have a scenario that runs react static where:

- The temporary directory is set to the OS tmp directory (i.e. `/tmp`)
- The output dist directory points to somewhere completely outside the subtree of the project
- The root directory of the project is not a sub-directory of the directory project containing react static

It also gets a bit weird when the package that includes react static is also a dependency (i.e. react static would be a dependency of a dependency of the project) and its installed globally, or in a `node_modules` directory that is not within the project.

SleeplessByte:
> I think the best way to test this is to have a scenario that runs react static where: [..]

Yes, I like this idea. I think this would be a good way to start our e2e integration test (which should do exactly this).

> The root directory of the project is not a sub-directory of the directory project containing react static
It also gets a bit weird when the package that includes react static is also a dependency 

Agreed. 

One of my plans for #1262 (which I didn't list there) is to separate the core from `@react-static/cli`. Then, we'll add a NEW package `@react-static/e2e` who's purpose is literally making sure the dependency graph isn't cyclic. We might also consider having `@react-static/templates`, but that's a different story.

SleeplessByte:
@jviotti can you update this one as well? Seems to add a few more resolve/relative fixes.

jviotti:
@SleeplessByte I rebased this one, and everything seems fine. Can you give it another look?

SleeplessByte:
I'm so excited for all these fixes @jviotti !

jviotti:
@SleeplessByte Great stuff! BTW, I'm very interested in getting react static to work with the root/tmp/dist directory customisations I described above. If you create an e2e test that attempts this, then I'm happy to help make it pass. I tried this myself in the past, and there were many things broken that prevented this. Some of them are the ones I fixed so far, but there are various more.

SleeplessByte:
Yeah, it's on the list. We as a company are _very_ busy for the next ~30-~35 days, but after that there should be much more speed. Right now I'm hotfixing everything and getting a release out, and then we'll test these properly!

jviotti:
Sounds like plan!

Issue #1143, opened 2019-04-21
[Bug] Unknown plugin hooks: "plugins" found in plugin


Herohtar:
The docs for [`node.api.js`](https://github.com/nozzle/react-static/blob/master/docs/plugins/node-api.md) state that you can use plugins in a plugin the same way you can in `static.config.js`; however, when I try to do so I get the error: `Unknown plugin hooks: "plugins" found in plugin`

### Environment

```
  System:
    OS: Windows 10
    CPU: (8) x64 AMD FX(tm)-8350 Eight-Core Processor
    Memory: 5.49 GB / 15.96 GB
  Binaries:
    Node: 11.10.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.13.0 - C:\Program Files\nodejs\yarn.CMD
    npm: 6.7.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: 44.17763.1.0
    Internet Explorer: 11.0.17763.1
  npmPackages:
    react: ^16.8.6 => 16.8.6
    react-dom: ^16.8.6 => 16.8.6
    react-jss: ^8.6.1 => 8.6.1
    react-markdown: ^4.0.8 => 4.0.8
    react-moment: ^0.9.2 => 0.9.2
    react-static: ^7.0.9 => 7.0.9
    react-static-plugin-jss: ^7.0.8 => 7.0.8
    react-static-plugin-reach-router: ^7.0.8 => 7.0.8
    react-static-plugin-sitemap: ^7.0.8 => 7.0.8
    react-static-plugin-source-filesystem: ^7.0.8 => 7.0.8
```

### Steps to Reproduce the problem

**Base your steps off of any freshly installed react-static template!**

1.  Create a `node.api.js` plugin
2.  Add any plugin to that plugin's config
3.  Build

### Reproducible Demo
https://github.com/Herohtar/thalion-herohtar/tree/jss-plugin

### Full error
```
> Executing task in folder thalion-herohtar: npm run stage <


> thalion-herohtar@0.1.0 stage C:\Users\Me\Desktop\HTML\thalion-herohtar
> react-static build --staging

Bundling application for Staging...

  Error: The following error occurred in the plugin: "C:\Users\Me\Desktop\HTML\thalion-herohtar"

  Error: Unknown plugin hooks: "plugins" found in plugin: C:\Users\Me\Desktop\HTML\thalion-herohtar
  - plugins.js:26 validatePlugin
[...]

PR #1264, opened 2019-07-17
Allow plugins to use plugins

Herohtar:
## Description

Yo dawg, I heard you like plugins, so I added 'plugins' to plugins.js so you can use plugins while you write plugins.

## Changes/Tasks

- [x] Added 'plugins' to the list of supported hooks for plugins

## Motivation and Context

Fixes https://github.com/react-static/react-static/issues/1143

## Types of changes

## Checklist:


SleeplessByte:
LGTM. I don't like that we don't test it, but then again, we don't really have good e2e testing right now.

PR #1283, opened 2019-08-14
🚀 7.2.0 maintenance

SleeplessByte:
- [x] Set-up workspace
- [ ] Document setting up workspace
- [x] Update minor and patch dependencies
- [x] Attempt dependency updates where they are breaking (major updates)
- [ ] Update version information, repository and author
- [ ] Update CHANGELOG.md
- [x] Test against test-suite
- [x] Test against production builds (locally)
- [ ] Update documentation where appropiate
- [ ] Publish to npm


SleeplessByte:
Note to self: Locally I have 2 failing tests, 

- [ ] One of which is a cross-env windows test (so the code is actually correct, but the test doesn't account for different pathing structures). 
- [ ] The other one seems to be related to updating `chokidar` past a major version. Downgrade it , re-run and then figure out if there is a path where we can upgrade past a major version.

SleeplessByte:
Note to self: since the hardcoded `overrideBrowserslist` is working, it can now be removed. `browserlists` will fall back to `defaults` if there is no config. It's technically breaking, but also a bug, so it will be released in this minor change.

- [ ] remove "hardcoded fix" from https://github.com/react-static/react-static/pull/1283/commits/7c244e45dbdbf8dd12cc4c09ad412253db869cf2, relying on auto-resolution of the config
- [ ] add documentation how to _resolve_ the mismatch between the old defaults and the new defaults, by telling users of `react-static` to add the config to their project.

SleeplessByte:
Note to self:

- [ ] investigate build on windows vs linux. Dev mode ís working, build seems to hang. Run with debug on. Probably minor cross env pathing thing.

SleeplessByte:
Note to self:

- [ ] can't seem to fix the _super weird_ chokidar issue. `path`, the node module, for some reason, does not have the right exports. Works fine when isolated in a different project.
- [ ] `--debug` build works. No errors! Why does it hang without this flag?

Pixelatex:
@SleeplessByte do you have access to travis? They seem to crash with every PR.. should be looked into.

SleeplessByte:
> @SleeplessByte do you have access to travis? They seem to crash with every PR.. should be looked into.

That's the chokidar issue. I have it locally too. I don't know yet _why_ it's caused in the first place.

Somewhere, for some reason, `path` is overwritten and doesn't return the node `path` module.

Pixelatex:
> 
> That's the chokidar issue. I have it locally too. I don't know yet _why_ it's caused in the first place.
> 
> Somewhere, for some reason, `path` is overwritten and doesn't return the node `path` module.

IF I have time this week, I'll try to help you debug this weird and see if I can get it working. Maybe a second pair of eyes will help.

SleeplessByte:
I would appreciate it @Pixelatex . My guess it's a load, babel or mock/jest issue...

nddery:
I don't know if this is new information or whatnot, but if I comment out the [following lines](https://github.com/react-static/react-static/blob/1bf18e07ad6df6f71a68ff0d0a39d2b7b7257f88/packages/react-static/src/static/__tests__/getConfig.test.js#L28-L32) I don't get the chokidar issue - but I do get new failures as paths aren't mocked correctly (expected).

I've seen this `undefined` issue with Jest when dealing with circular dependencies, but don't think this is happening here.

SleeplessByte:
It is! And actually exactly matches my findings. I updated jest as well and these mocks are wrong anyway. 

I'll rework that to fix chokidar, now the hanging builds without the dev flag...

alexjfno1:
@SleeplessByte how are you getting on with this? If you can give me the steps to reproduce the hanging issue I can take a look?

SleeplessByte:
I'm on windows. @nddery and I think it might be a windows issue.

@alexjfno1 

- Running `build` hangs for me
- Running `build --debug` doesn't hang for me

Perhaps it's one of the dependencies! When running _without_ `--debug` and if `NODE_ENV` is NOT yet set, it will be set to `production`. Internally we use `REACT_STATIC_ENV`, so it doesn't matter for the react-static parts. Perhaps flicking on production is the actual issue.

SleeplessByte:
If I don't figure this out tonight, I'm just going to release it with a note. I'll then work on 8.0.0 and fix these cross-env things "from scratch".

alexjfno1:
@SleeplessByte I've pulled your branch and I don't see the issue.

1. Pulled branch and ran `yarn` and `yarn build`
2. `./packages/react-static/bin/react-static create` and made a basic project
3. `cd ./that-project` 
4. `yarn build` was successful apart from some warnings about autoprefixer and asset sizes

Using OSX High Sierra, Node `10.15.3` and NPM `6.9.0`

Maybe a Windows issue?

SleeplessByte:
Thank you for confirming. It is likely. I have Max, linux and windows here, so I am going to try it everywhere, confirm, release and then work on a fix.

SleeplessByte:
Closed in favour of #1292

PR #1326, opened 2019-10-11
Updated docs about `Routes` component

dallonf:
## Description

Rewrote some of the docs regarding the `Routes` component, particularly to correct language and examples suggesting that you can use the children-as-function pattern with it.

## Changes/Tasks

- [x] Changed API docs
- [x] Changed `animated-routes` example

## Motivation and Context

In trying to implement this component's advanced features in my own site, I found that the docs were misleading and in some cases just wrong. The names are still somewhat confusing (why is it called `getComponentForPath` if it returns an element, not a component?) but at least the information is now correct (as far as I can tell from looking through the source).

I also updated the `animated-routes` example the best I could, but I'm not really familiar with `react-spring` and there are still some pretty big issues in the example. Namely, the temporary routes kind of stack on top of each other while transitioning, creating a janky transition, and the idea of temporarily rendering a component that isn't the current one isn't really compatible with the `useRouteData()` hook (try implementing this example in the "basic" template and navigate away from the Blogs page).

## Types of changes

## Checklist:

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -38,31 +38,30 @@ The routes that will be rendered are the **routes** returned by the `getRoutes` 
 
 ### Custom `Routes` Rendering
 
-Occasionally, you may need to render the automatic `<Routes>` component in a custom way. The most common use-case is illustrated in the [animated-routes](https://github.com/react-static/react-static/tree/master/examples/animated-routes) example transitions. To do this, utilize a render prop:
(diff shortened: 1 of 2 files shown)


dallonf:
Oops, my bad, didn't realize my Prettier config was messing with the style! How's this?

SleeplessByte:
> why is it called getComponentForPath if it returns an element, not a component?

Legacy, we can change it in RS 8+ :)

Issue #1303, opened 2019-09-06
[Bug] Absolute paths leaking in the bundle


nddery:
### Reporting a bug?

Absolute paths are found in the exported bundles. This is surprising, can leak information about your build system, and will cause [webpack longterm caching](https://webpack.js.org/guides/caching/) to not work if you build your site in different directories over time (or on multiple servers).

I believe the caching is two fold, part of it would be to use [`contenthash`](https://github.com/react-static/react-static/pull/1232) and second would be to use relative path for referencing modules.

I know there's been a lot of work put into using absolute paths over relative paths lately, and I wonder if we could achieve the same result with relative paths instead (unclear what the issue was) ?

### Environment

<details>
<summary>envinfo result</summary>

```
❯ npx envinfo --system --npmPackages 'react*' --binaries --npmGlobalPackages 'react*' --browsers
npx: installed 1 in 1.434s

  System:
    OS: macOS 10.14.5
    CPU: (4) x64 Intel(R) Core(TM) i7-7567U CPU @ 3.50GHz
    Memory: 6.09 GB / 16.00 GB
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 8.14.0 - ~/n/bin/node
    Yarn: 1.16.0 - /usr/local/bin/yarn
    npm: 6.9.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 76.0.3809.132
    Firefox: 69.0
    Safari: 12.1.1
  npmPackages:
    react: ^16.9.0 => 16.9.0
    react-dom: ^16.9.0 => 16.9.0
    react-static: ^7.2.0 => 7.2.2
    react-static-plugin-reach-router: ^7.2.0 => 7.2.2
    react-static-plugin-sitemap: ^7.2.0 => 7.2.2
    react-static-plugin-source-filesystem: ^7.2.0 => 7.2.2
```
</details>

### Steps to Reproduce the problem

1. Install a "basic" react-static site
```
❯ react-static create

? What should we name this project? my-static-site
? Select a template below... basic
Creating new react-static project...
```

2. `cd my-static-site`
3. `yarn build`
4. Look in the `dist/main.*.js` file and find absolute paths. If you're on Mac, this mean you'd find `/Users/[your-user]/...` paths in there.

Another way to see these absolute path is to run `yarn serve` and look at the `sources` tab in the devtools. You should see absolute path there: 
[...]

PR #1329, opened 2019-10-15
Use assetsPath as Webpack publicPath

dallonf:
## Changes/Tasks

- [x] Changed `output.publicPath` to the `REACT_STATIC_ASSETS_PATH ` variable

## Motivation and Context

Based on what `publicPath` [does in Webpack](https://webpack.js.org/configuration/output/#outputpublicpath), I believe it's actually better set to what React-Static calls `assetsPath`. The dev Webpack configuration already works this way.

In our use case, the assets are hosted on a different CDN than the actual site - if the `siteRoot` is `https://www.example.com`, the `assetRoot` is `https://xxxxx.cloudfront.net`. However, certain imports (mostly images imported like `import imgSrc from '../img.png';`) were still resulting in a path like `https://www.example.com/static/img.png`.

## Types of changes

## Checklist:

- [ ] I have updated the documentation accordingly*
- [ ] My changes have tests around them**

*This is just a bugfix, I don't think documentation changes are necessary
** Don't think there's anything to test here

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 - Fix "can not read property `catch` of `undefined`" ([#1313](https://github.com/react-static/react-static/pull/1313))
 - Fix missing state.siteData in dev ([#1148](https://github.com/react-static/react-static/pull/1148))
 - Add clickable dev-server url ([#1306](https://github.com/react-static/react-static/pull/1306))
-- Fix `output.publicPath` in dev build when `assetPath` is set ([#1329](https://github.com/react-static/react-static/pull/1329))
+- Fix `output.publicPath` in prod build when `assetPath` is set ([#1329](https://github.com/react-static/react-static/pull/1329))
 
 ## 7.2.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@
 - Fix missing state.siteData in dev ([#1148](https://github.com/react-static/react-static/pull/1148))
 - Add clickable dev-server url ([#1306](https://github.com/react-static/react-static/pull/1306))
 
+### Fixes & Optimizations
+
+- Fix `output.publicPath` in dev build when `assetPath` is set ([#1329](https://github.com/react-static/react-static/pull/1329))
+
 ## 7.2.2
 
 ### Improved
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,9 +8,6 @@
 - Fix "can not read property `catch` of `undefined`" ([#1313](https://github.com/react-static/react-static/pull/1313))
 - Fix missing state.siteData in dev ([#1148](https://github.com/react-static/react-static/pull/1148))
 - Add clickable dev-server url ([#1306](https://github.com/react-static/react-static/pull/1306))
-
-### Fixes & Optimizations
-
 - Fix `output.publicPath` in dev build when `assetPath` is set ([#1329](https://github.com/react-static/react-static/pull/1329))
 
 ## 7.2.2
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,10 @@
 - Fix missing state.siteData in dev ([#1148](https://github.com/react-static/react-static/pull/1148))
 - Add clickable dev-server url ([#1306](https://github.com/react-static/react-static/pull/1306))
 
+### Fixes & Optimizations
```
SleeplessByte: This would fall under `improved`

SleeplessByte (approved):
Can you just remove the heading?

SleeplessByte:
Perfect!

cc @nddery @jviotti 

If you have a spare few minutes, would this change work for your set-ups?

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 - Fix "can not read property `catch` of `undefined`" ([#1313](https://github.com/react-static/react-static/pull/1313))
 - Fix missing state.siteData in dev ([#1148](https://github.com/react-static/react-static/pull/1148))
 - Add clickable dev-server url ([#1306](https://github.com/react-static/react-static/pull/1306))
+- Fix `output.publicPath` in dev build when `assetPath` is set ([#1329](https://github.com/react-static/react-static/pull/1329))
```
nddery: Shouldn't this be "in prod build" ?

nddery (commented):
I'm thinking this might lead to issues when `config.assetsPath` is not absolute (ie.: when it's set to "assets", for example) and when the user sets `config.basePath`. Because of https://github.com/react-static/react-static/blob/master/packages/react-static/src/static/getConfig.js#L141-L143 the asset path would be wrong no ?

This also seems to be breaking when only a `basePath` is set, because assets being prefetched do not include the base path.

CHANGELOG.md
dallonf: Yes... yes it should. Oops

dallonf:
> I'm thinking this might lead to issues when `config.assetsPath` is not absolute (ie.: when it's set to "assets", for example) and when the user sets `config.basePath`. Because of https://github.com/react-static/react-static/blob/master/packages/react-static/src/static/getConfig.js#L141-L143 the asset path would be wrong no ?
> 
> This also seems to be breaking when only a `basePath` is set, because assets being prefetched do not include the base path.

This sounds like the issue is a lot more complicated than anticipated. I'm not sure of what the real solution is. Should I close this PR and open a bug ticket instead describing the problem?

SleeplessByte:
Most likely it "just" needs a check to see if those bases are set, and if not, "fix" the path resolution. There are various bugs currently due to switching to absolute paths and I am looking into "undoing it", by means of supporting absolute paths only in specific places.

Anyway, both @nddery and @jviotti have build systems currently where they build in a different "tree" than their project lives (if I recall correctly), and they've both submitted various PRs to make that work. If your changes work with their systems, and we can capture edge cases, this is fine to merge. Will be changed in RS 8, which is weeks to months out.

nddery:
We've configured our react-static to build in the default `dist` folder. We do however use the `basePath` configuration, as well as `assetsPath` (set to "assets" - same value as `paths.assets`). Being able to set an `assetsPath` that points to a CDN would, I think, be useful for us, too. 

The absolute paths issue is a blocker for us to upgrade to v7.2.2, which I've been trying to address in https://github.com/react-static/react-static/issues/1303 (but have not been able to put a lot of time into this lately...). I'd be interested to hear more about your thoughts on "supporting absolute paths only in specific places" (which might be a different topic than this PR)!

My biggest hurdle is the shear number of different configuration possible, which I find to be the same issue here, a simple change can have big impact on different configurations. I've been thinking that a thorough review of every configuration options and how they interplay with each other might help.

SleeplessByte:
> My biggest hurdle is the shear number of different configuration possible, which I find to be the same issue here, a simple change can have big impact on different configurations. I've been thinking that a thorough review of every configuration options and how they interplay with each other might help.

Perhaps we can slash a few?

kvet:
@SleeplessByte will this ever be released? We are patching `react-static` in order to get it working

SleeplessByte:
I don't think it's currently finished

PR #1340, opened 2019-11-21
Remove absolute paths from generated JS

nddery:
Attempts to fix https://github.com/react-static/react-static/issues/1303.

## Description

Absolute paths are found in the exported bundles. This is surprising, can leak information about your build system, and will cause [webpack longterm caching](https://webpack.js.org/guides/caching/) to not work if you build your site in different directories over time (or on multiple servers).

(copy of https://github.com/react-static/react-static/issues/1303#issuecomment-535139590) I've tested changing both `paths.dist` and `paths.temp` to point outside of the working directory. This works fine and no absolute paths end up in the generated files, ensuring consistent hashing even when the build directory changes.

However, I can't get `yarn build` to successfully run when also changing `paths.buildArtifacts` to a directory that's outside the working directory. I've got the same issue when using the upstream `react-static@7.2.2`.

```
Exporting HTML...
  Error: Trace: { Error: Cannot find module '@reach/router'
      at Function.Module._resolveFilename (internal/modules/cjs/loader.js:636:15)
      at Function.Module._load (internal/modules/cjs/loader.js:562:25)
      at Module.require (internal/modules/cjs/loader.js:692:17)
      at Module.call [as require] (/react-static/packages/react-static/s  rc/utils/binHelper.js:62:26)
      at require (internal/modules/cjs/helpers.js:25:18)
      [...]
```

Unclear to me if this is a local setup issue with `yarn link` or if this would also manifest in a "real-world setup".

## Changes/Tasks

- [x] Remove usage of `REACT_STATIC_UNIVERSAL_PATH` in favor of webpack alias for it

At the time this was added (7363c0b) there were no alias set for `react-universal-component` set in the Webpack config. Now that there is, this is redundant as Webpack will automatically resolve to our chosen module.

- [x] Ensure HMR code isn't found in production bundle

This was including absolute paths, which was breaking long term caching. Unclear why this isn't picked up by Terser and removed during dead-code elimination ?

- [x] Adds a `__react_static_root__` webpack alias

This ensure we do not output absolute paths for templates and browsers plugins.

- [x] Use `contenthash` over `hash` and `chunkhash`

According to Webpack documentation, `hash` is unique per build. When building on multiple machines, this hash is different on all builds. Using `contenthash` ensure we get deterministic hashes between multiple machines and between builds.
https://webpack.js.org/configuration/output/#outputfilename

Webpack also recommend using `contenthash` for long term caching.
https://webpack.js.org/guides/caching/#output-filenames

## Types of changes

## Checklist:

### DIFF
--- a/packages/react-static/src/static/__tests__/extractTemplates.test.js
+++ b/packages/react-static/src/static/__tests__/extractTemplates.test.js
@@ -46,26 +46,30 @@ test('the 404 template is the first one', async () => {
   expect(state.templates[0]).toContain('src/templates/404')
 })
 
-test('relative routes path resolves against the ROOT', async () => {
+test('relative routes path are relative and use the __react_static_root__ alias', async () => {
   const { templates } = await extractTemplates({
     config,
     routes: [{ path: '404', template: './src/templates/NotFound' }],
   })
 
-  expect(path.isAbsolute(templates[0])).toBe(true)
-  expect(path.normalize(templates[0])).toBe(
-    path.resolve(`${config.paths.ROOT}/src/templates/NotFound`)
+  expect(templates[0]).toBe(
+    `__react_static_root__/${path.relative(
+      config.paths.ROOT,
+      './src/templates/NotFound'
+    )}`
   )
 })
 
-test('absolute routes path are kept intact', async () => {
+test('absolute routes path are relative and use the __react_static_root__ alias', async () => {
   const { templates } = await extractTemplates({
     config,
     routes: [{ path: '404', template: '/home/src/templates/NotFound' }],
   })
 
-  expect(path.isAbsolute(templates[0])).toBe(true)
-  expect(path.normalize(templates[0])).toBe(
-    path.resolve('/home/src/templates/NotFound')
+  expect(templates[0]).toBe(
+    `__react_static_root__/${path.relative(
+      config.paths.ROOT,
+      '/home/src/templates/NotFound'
+    )}`
   )
 })
--- a/packages/react-static/src/bootstrapPlugins.js
+++ b/packages/react-static/src/bootstrapPlugins.js
@@ -5,7 +5,12 @@ const { registerPlugins } = require('./browser')
 
 registerPlugins(plugins)
 
-if (typeof document !== 'undefined' && module && module.hot) {
+if (
+  process.env.NODE_ENV !== 'production' &&
+  typeof document !== 'undefined' &&
+  module &&
+  module.hot
+) {
   module.hot.accept(process.env.REACT_STATIC_PLUGINS_PATH, () => {
     registerPlugins(require(process.env.REACT_STATIC_PLUGINS_PATH).default)
   })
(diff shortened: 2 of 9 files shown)


nddery:
(sorry, accidentally hit the "submit PR" button too soon.. updated the PR description and will look at the failing specs)

nddery:
@jviotti does this work with your setup ? I've tested this with both the default config for paths and pointing `paths.temp` and `paths.dist` to a directory outside of the current working directory.

packages/react-static/src/static/__tests__/extractTemplates.test.js
```diff
@@ -46,26 +46,30 @@ test('the 404 template is the first one', async () => {
...
   const { templates } = await extractTemplates({
     config,
     routes: [{ path: '404', template: '/home/src/templates/NotFound' }],
   })
 
-  expect(path.isAbsolute(templates[0])).toBe(true)
-  expect(path.normalize(templates[0])).toBe(
-    path.resolve('/home/src/templates/NotFound')
+  expect(templates[0]).toBe(
+    `__react_static_root__/${path.relative(
```
nddery: Just a note that this would produce URLs like this, where `__react_static_root__` is part of the folder structure and the URL:

- http://localhost:5000/templates/__react_static_root__/src/containers/Post.f46e6814.js

SleeplessByte:
@nddery ping me when I need to look at this!

nddery:
@SleeplessByte this would be ready for review, as mentioned in the description, I couldn't get to update the `paths.buildArtifacts` config but the rest should work. Unclear if this is an issue with my local `yarn link` setup or not, but I get the same results without `yarn link` (using the package on the registry).

SleeplessByte:
It might be how webpack is configured (as in buildArtifacts not in the resolution list). 

Anyway, will review this. Thank you so much for tackling this. I'm obviously using your work for RS 8 as well, making sure the chunking/hashing is correct there.

SleeplessByte:
People who care about this in RS 8+: https://spectrum.chat/react-static/rfcs/rfc-import-path-aliases~6a221f30-060e-4c7a-887b-b8f65fb86319

@nddery sofar this looks good. I'll do a bit more testing and then merge.

nddery:
@SleeplessByte re: the RFC you linked above, my understanding is that the first option you've outlined is what react-static used to output, before all the absolute paths changes - would you be able to get the relative paths working without breaking projects with configuration like @jviotti  ?

SleeplessByte:
You got it.

I basically was trying to figure out where we need the absolute paths and the answer is nowhere (re: our previous discussion).

However, if the src and dist root are far away, the path between them may leak information. So I will add the alias you've come up with, especially in that case.

My question is:
Do we want it always it only when it's like that... Far apart dirs.

SleeplessByte:
By the way, I feel like merging this in the next week or so regardless; unless something is mentioned by people tagged in the messages above.

nddery:
> Do we want it always it only when it's like that... Far apart dirs.

I tend to favor not using aliases like that - but also tend to prefer having a single way of doing something. This said, if we're to sometimes use the alias, I would favor always using it.

SleeplessByte:
We are moving offices the coming 10 days, but will try to release this intermittently.

PR #1348, opened 2019-12-23
Pass options to css-loader

just-paja:
Pass options from sass plugin to css-loader to enable css modules.

## Description

Add `cssLoaderOptions` property to enable configuration of css-loader.

## Changes/Tasks

- [✓] Changed code

## Motivation and Context

Enables developers the CSS modules feature by providing options to configure css-loader used inside the react-static-plugin-sass

## Types of changes

- [ - ] Refactoring/add tests (refactoring or adding test which isn't a fix or add a feature)
- [ - ] Bug fix (non-breaking change which fixes an issue)
- [✓] New feature (non-breaking change which adds functionality)
- [ - ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [✓] I have updated the documentation accordingly
- [✓] I have updated the CHANGELOG with a summary of my changes
- [ - ] My changes have tests around them

### DIFF
--- a/packages/react-static-plugin-less/src/node.api.js
+++ b/packages/react-static-plugin-less/src/node.api.js
@@ -3,7 +3,7 @@ import autoprefixer from 'autoprefixer'
 import postcssFlexbugsFixes from 'postcss-flexbugs-fixes'
 import semver from 'semver'
 
-export default ({ includePaths = [], ...rest }) => ({
+export default ({ includePaths = [], cssLoaderOptions, ...rest }) => ({
   webpack: (config, { stage }) => {
     let loaders = []
     const lessLoaderPath = require.resolve('less-loader')
@@ -17,6 +17,7 @@ export default ({ includePaths = [], ...rest }) => ({
       options: {
         importLoaders: 1,
         sourceMap: false,
+        ...cssLoaderOptions,
       },
     }
     const postCssLoader = {
--- a/packages/react-static-plugin-less/src/node.api.js
+++ b/packages/react-static-plugin-less/src/node.api.js
@@ -3,7 +3,7 @@ import autoprefixer from 'autoprefixer'
 import postcssFlexbugsFixes from 'postcss-flexbugs-fixes'
 import semver from 'semver'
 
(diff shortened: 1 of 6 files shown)


just-paja:
Just noticed that the same config is [already used with react-static-plugin-stylus](https://github.com/react-static/react-static/blob/master/packages/react-static-plugin-stylus/src/node.api.js#L24). Looks like we're thinking the same :-) I will extend this PR to do the same for [react-static-plugin-less](https://github.com/react-static/react-static/tree/master/packages/react-static-plugin-less).

packages/react-static-plugin-less/src/node.api.js
```diff
@@ -3,7 +3,7 @@ import autoprefixer from 'autoprefixer'
 import postcssFlexbugsFixes from 'postcss-flexbugs-fixes'
 import semver from 'semver'
 
-export default ({ includePaths = [], ...rest }) => ({
+export default ({ includePaths = [], cssLoaderOptions, ...rest }) => ({
```
SleeplessByte:
```suggestion
export default ({ includePaths = [], cssLoaderOptions = {}, ...rest }) => ({
```

Probably want to default to an empty object?

packages/react-static-plugin-sass/src/node.api.js
```diff
@@ -3,7 +3,7 @@ import autoprefixer from 'autoprefixer'
 import postcssFlexbugsFixes from 'postcss-flexbugs-fixes'
 import semver from 'semver'
 
-export default ({ includePaths = [], ...rest }) => ({
+export default ({ includePaths = [], cssLoaderOptions, ...rest }) => ({
```
SleeplessByte:
```suggestion
export default ({ includePaths = [], cssLoaderOptions = {}, ...rest }) => ({
```

just-paja:
Applied the suggested changes and updated the plugin-stylus code to look the same.

just-paja:
@SleeplessByte , thank you for merging it and sorry I did not have more time to invest into your library. I tested passing the options using local build with linked version. Not sure if I would be able to write a meaningful test that verifies this functionality.

Is there any plan to release this?

PR #1356, opened 2020-01-07
Progress module writes on stderr, instead of stdout.

godjirax:
Progress module writes on stderr, instead of stdout.

## Description

I changed the output from stderr to stdout.

## Motivation and Context

I use `rush` (https://rushjs.io/) to handle a monorepo.
It looks for stderr output to check if there are errors during the build step.
If there's an error, it stops the build pipeline.

## Types of changes

## Checklist:


SleeplessByte:
> If there's an error, it stops the build pipeline.

Should probably use the exit code.

Anyway, this is a fine change.

godjirax:
> > If there's an error, it stops the build pipeline.
> 
> Should probably use the exit code.
> 

They do use it also.
If something is written to the stderr, they marked as a warning which changes the exit code than 0.

I will update the changelog.

PR #1373, opened 2020-02-05
Add inline script hashes to DocumentProps

### LINKED ISSUE #1362: Make site work without `unsafe-inline` CSP
**Is your feature request related to a problem? Please describe.**

I've been setting up a Content Security Policy (CSP) for a react-static site and I've noticed that the way static html is generated currently requires `script-src 'unsafe-inline'` and `style-src 'unsafe-inline';` policies. Adding these policies opens up XSS and other attack vectors, so I'd prefer if the site worked without these policies.

Looking at a `react-static-example-basic` site, there are two issues. One is an inline script with the route info:

```html
<script type="text/javascript">
    window.__routeInfo = JSON.parse("{\"template\":\"READACTED/my-static-site/src/pages/index.js\",\"sharedHashesByProp\":{},\"data\":{},\"path\":\"/\",\"sharedData\":{},\"siteData\":{}}");
</script>
```

And two divs with the following inline style:

```html
<div style="outline:none" tabindex="-1" role="group">
```

**Describe the solution you'd like**

It would be great if there was an option to not to include the inline script and load the route info from the `routeInfo.json` at the cost of performance. I'm not sure where the `outline:none` style comes from and why it's necessary, but it would be great if that could be excluded somehow as well.

**Describe alternatives you've considered**

For the inline script, one could compute its hash and insert a meta tag with the CSP option  `script-src '<hash-algorithm>-<base64-value>';`, but I'm not sure how this would play with additional CSP configurations eg. from headers.

Happy to provide a PR with some guidance :)

abiro:
## Description

This PR adds a new object to the `state` attribute of `DocumentProps` called `inlineScripts`. The keys of the `inlineScripts` object are descriptive names of the scripts and contain the script source and its SHA256 hash. There is currently only one inline script, the route info hydration script:

```json
  { 
      "routeInfo": { 
          "script": "window.__routeInfo = JSON.parse(\"route data as json string\")", 
          "hash": "sha256-<base64-value>",
      }
  }
```

These hashes can be used to construct a Content Security Policy in a meta tag without needing `unsafe-inline` for scripts. 

## Changes/Tasks

- [x] Refactor `generateRouteInformation` and move it from `BodyWithMeta.js` to `exportRoute.js`, as the inline route info script hash has to be computed there.
- [x] Move testing inline route info script from `BodyWithMeta.test.js` to `exportRoute.test.js`
- [x] Update `BodyWithMeta.test.js.snap` to only check for presence of injected script, not the script's contents, since the script isn't generate in `BodyWithMeta` anymore.
- [x] Compute SHA256 hash for route info inline script in `exportRoute`.
- [x] Test script hashes in `exportRoute.test.js`
- [x] Add `inlineScripts` member to `state` in `exportRoute` that holds the script source and its hash
- [x] Add `inlineScripts` to to `DocumentProps` interface in the type def file.
- [x] Add `inlineScripts` to `Document` documentation in `docs/config.md`.

## Motivation and Context

Resolves #1362

## Types of changes

- [x] Refactoring/add tests (refactoring or adding test which isn't a fix or add a feature)

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Add unofficial plugin `react-static-plugin-file-watch-reload` to plugins list
 - Enable configuring css loader from `react-static-plugin-sass` and `react-static-plugin-less` ([#1348](https://github.com/react-static/react-static/pull/1348))
 - Update `react-static-plugin-jss` for react-jss v10+. ([#1367](https://github.com/react-static/react-static/pull/1367))
+- Add inline script hashes to `DocumentProps`. These hashes can be used to construct a Content Security Policy in a meta tag without `unsafe-inline` scripts.
 
 
 ### Improved
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 - Add unofficial plugin `react-static-plugin-file-watch-reload` to plugins list
 - Enable configuring css loader from `react-static-plugin-sass` and `react-static-plugin-less` ([#1348](https://github.com/react-static/react-static/pull/1348))
 - Update `react-static-plugin-jss` for react-jss v10+. ([#1367](https://github.com/react-static/react-static/pull/1367))
-- Add inline script hashes to `DocumentProps`. These hashes can be used to construct a Content Security Policy in a meta tag without `unsafe-inline` scripts.
+- Add inline script hashes to `DocumentProps`. These hashes can be used to construct a Content Security Policy in a meta tag without `unsafe-inline` scripts. ([#1373](https://github.com/react-static/react-static/pull/1373))
 
 
 ### Improved
--- a/packages/react-static/src/index.d.ts
+++ b/packages/react-static/src/index.d.ts
@@ -584,6 +584,22 @@ export interface PathsConfig {
 
 export type DocumentComponent<T extends object = any> = React.ComponentType<DocumentProps<T>>
 
+export interface InlineScript {
+  /**
+   * The script as a string.
+   */
+  script: string
+  /**
+   * The script's sha256 hash in base64 using the subresource integrity format,
+   * eg.: 'sha256-<base64-value>'. This value can be directly used in a CSP.
+   */
+  hash: string
+}
+
+export interface InlineScripts {
+  routeInfo: InlineScript
+}
+
 export interface DocumentProps<T extends object = any> {
   Html: JSX.IntrinsicElements['html']
   Head: JSX.IntrinsicElements['head']
@@ -592,7 +608,8 @@ export interface DocumentProps<T extends object = any> {
   state: {
     siteData: T
     routeInfo: unknown
-    renderMeta: unknown
+    renderMeta: unknown,
+    inlineScripts: InlineScripts
   }
 }
 
--- a/packages/react-static/src/index.d.ts
+++ b/packages/react-static/src/index.d.ts
@@ -596,7 +596,7 @@ export interface InlineScript {
   hash: string
 }
 
-export interface InlineScripts {
+export interface InlineScripts extends Record<string, InlineScript> {
   routeInfo: InlineScript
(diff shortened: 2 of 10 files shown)


abiro:
There is an unrelated [change in the commits](https://github.com/react-static/react-static/pull/1373/commits/d482bdb5fba51658b21d8bc17d956a2901b8c1b2) that prevents git from rewriting line endings in PNG files. I had to add this, because the `* text eol=lf` in `.gitattributes` was making git rewrite PNG files in the repo on Ubuntu. Let me know please if I should break this out to a different PR.

abiro:
I also tried to test this out in my project by doing a `yarn pack` in `packages/react-static` and then installing that with `npm install file:/...`, but then `react-static start` failed with

```
Trace: Error: Cannot find module '../lib/utils/binHelper'
```

I'm wondering if you can suggest a better workflow for local installations?

SleeplessByte:
> I'm wondering if you can suggest a better workflow for local installations?

```
yarn build
yarn link
```


then in a new project `yarn link react-static`

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 - Add unofficial plugin `react-static-plugin-file-watch-reload` to plugins list
 - Enable configuring css loader from `react-static-plugin-sass` and `react-static-plugin-less` ([#1348](https://github.com/react-static/react-static/pull/1348))
 - Update `react-static-plugin-jss` for react-jss v10+. ([#1367](https://github.com/react-static/react-static/pull/1367))
+- Add inline script hashes to `DocumentProps`. These hashes can be used to construct a Content Security Policy in a meta tag without `unsafe-inline` scripts.
```
SleeplessByte: Add the PR link please :)

packages/react-static/src/index.d.ts
```diff
@@ -584,6 +584,22 @@ export interface PathsConfig {
...
+   */
+  script: string
+  /**
+   * The script's sha256 hash in base64 using the subresource integrity format,
+   * eg.: 'sha256-<base64-value>'. This value can be directly used in a CSP.
+   */
+  hash: string
+}
+
+export interface InlineScripts {
```
SleeplessByte: We probably want "generic object, including this one":

```typescript
export interface InlineScripts extends Record<string, InlineScript> {
   routeInfo: InlineScript
}
```

This _should_ allow us to do this:

```typescript
const scripts: InlineScripts = {
  routeInfo: { script: '...', hash: '...' }
}
// this should work

scripts['anything'] = { script: '...', hash: '...' }
// this should work

scripts.
// this should autocomplete `routeInfo`
```

It then _disallows_:

```typescript
const scripts: InlineScripts = {
  routeInfo: { script: '...', hash: '...' }
}
scripts['foo'] = 'bar'
// Type '"bar"' is not assignable to type InlineScript
```

CHANGELOG.md
abiro: Oops, done in https://github.com/react-static/react-static/pull/1373/commits/fa8d0c00f56e006405df494188fb308e9042d986

packages/react-static/src/index.d.ts
abiro: Nice, added in https://github.com/react-static/react-static/pull/1373/commits/d76638ea10606be4e10d110883f449f9865f970c

abiro:
> > I'm wondering if you can suggest a better workflow for local installations?
> 
> ```
> yarn build
> yarn link
> ```
> 
> then in a new project `yarn link react-static`.

Thanks, I've tested it out locally now and the feature works as expected.

(Added these instructions to CONTRIBUTING in https://github.com/react-static/react-static/pull/1373/commits/7507d07bb1de659674c7f66020a609be8762a28f)

abiro:
Thanks for the quick review/merge @SleeplessByte :)

Issue #1360, opened 2020-01-11
[Bug] Error: EINVAL: invalid argument, mkdir 'E:\www\test-react-static\dist\templates\E:\www\test-react-static\src\pages


chiqui3d:
### Reporting a bug?
Error: EINVAL: invalid argument, mkdir 'E:\www\test-react-static\dist\templates\E:\www\test-react-static\src\pages

### Environment

`npx envinfo --system --npmPackages react* --binaries --npmGlobalPackages react* --browsers` 

here:

```

  System:
    OS: Windows 10 10.0.19041
    CPU: (6) x64 AMD Phenom(tm) II X6 1090T Processor
    Memory: 659.62 MB / 12.00 GB
  Binaries:
    Node: 12.13.0 - C:\Program Files\nodejs\node.EXE
    npm: 6.13.6 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: 44.19041.1.0
    Internet Explorer: 11.0.19041.1
  npmPackages:
    react: ^16.9.0 => 16.12.0
    react-dom: ^16.9.0 => 16.12.0
    react-helmet: ^5.2.1 => 5.2.1
    react-static: ^7.2.0 => 7.2.3
    react-static-plugin-reach-router: ^7.2.0 => 7.2.3
    react-static-plugin-sitemap: ^7.2.0 => 7.2.3
    react-static-plugin-source-filesystem: ^7.2.0 => 7.2.3

```

of course include any other package versions here if relevant. 

### Steps to Reproduce the problem

1.  react-static build

SleeplessByte:
Can you try to downgrade to 7.1.0? I don't know what's going on here. Which template did you use? Have you configured anything in static.js?

chiqui3d:
Just install from **Basic** template. With version 7.1 I can build, but I get the following message:

```
 Replace Autoprefixer browsers option to Browserslist config.
  Use browserslist key in package.json or .browserslistrc file.

  Using browsers option cause some error. Browserslist config 
  can be used for Babel, Autoprefixer, postcss-normalize and other tools.

  If you really need to use option, rename it to overrideBrowserslist.

  Learn more at:
  https://github.com/browserslist/browserslist#readme
  https://twitter.com/browserslist



  Replace Autoprefixer browsers option to Browserslist config.
  Use browserslist key in package.json or .browserslistrc file.

  Using browsers option cause some error. Browserslist config 
[...]

PR #1375, opened 2020-02-06
🚀 Release 7.3.0

### LINKED ISSUE #1372: [Bug] Bundle not working
Bundling isn't working.

When it gets to `Bundling App...` is sits there forever. If I change the script to `react-static build --debug` it will build it successfully, but the html isn't minified.

### Environment

```
  System:
    OS: Windows 10 10.0.18362
    CPU: (12) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
    Memory: 6.70 GB / 15.90 GB
  Binaries:
    Node: 10.16.3 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.21.1 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 6.9.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: 44.18362.449.0
    Internet Explorer: 11.0.18362.1
  npmPackages:
    react: ^16.9.0 => 16.12.0
    react-dom: ^16.9.0 => 16.12.0
    react-router: ^5.1.2 => 5.1.2
    react-router-dom: ^5.1.2 => 5.1.2
    react-static: ^7.2.0 => 7.2.3
    react-static-plugin-react-router: ^7.2.3 => 7.2.3
    react-static-plugin-sass: ^7.2.2 => 7.2.2
    react-static-plugin-sitemap: ^7.2.0 => 7.2.3
    react-static-plugin-source-filesystem: ^7.2.0 => 7.2.3
```

1.  Clone `git@github.com:JoshClose/CsvHelper.git`
1.  Go into directory `CsvHelper/docs-src/docs`
1.  The project is currently linked to `react-static` source, so you'll need to `yarn link` `react-static` in the `react-static` source and then in this repo.
1.  Run `yarn`
1.  Run `yarn build`

### Expected Behavior

Building should not hang indefinitely.

SleeplessByte:
## Types of changes

- [X] Refactoring/add tests (refactoring or adding test which isn't a fix or add a feature)
- [X] New feature (non-breaking change which adds functionality)

## Checklist:

- [X] I have updated the documentation accordingly
- [X] My changes have tests around them

Fixes #1372 (I think -- doesn't change paths, just the hanging)

@nddery I don't think the paths generated by `extractTemplates` are necessary correct -- but alas.


SleeplessByte:
I'm just going to test the templates with this and then I'll release it.

sochix:
@SleeplessByte awaiting new release !

nddery:
> I don't think the paths generated by extractTemplates are necessary correct -- but alas.

@SleeplessByte is the `path.posix` change here fixing or attempting to fix them ? What is wrong with them ?

SleeplessByte:
Inside react static we have Linux paths (that's what the slash function is doing). I change here to POSIX because otherwise the test generates a windows path and the internal creates a POSIX path.

The value itself is what worries me. It's something like:

`__react_static_root__/../../../../../../src/path/to/whatever`

nddery:
> The value itself is what worries me. It's something like:
>
> __react_static_root__/../../../../../../src/path/to/whatever

Yeah.. that is unexpected. `__react_static_root__` should have replaced all those `../../../`.

Pulling this branch locally and building the "basic" template, paths toward the templates look like: `__react_static_root__/src/path/to/...`.  With what configuration (mainly `paths.*`) are you getting these results ?

SleeplessByte:
> With what configuration (mainly paths.*) are you getting these results ?

I'm seeing it in the _tests_ that I changed (which were failing on windows, but not on linux). I'm not seeing it anywhere else so far :)
I'm wondering if this is related to #1360? in that case it puts two absolute paths together haha

nddery:
I wonder if it would make sense to also run the test suite on Windows - Travis seems to support "Windows Server, version 1809".  I'm not sure what would cause this and unfortunately don't have a Windows machine to debug further.

SleeplessByte:
I probably will move the testing away from Travis this month. It's been failing intermittently and GH Actions work just fine. That said, I have more than one windows runner and so do my co-workers so we'll be alright.

barroudjo:
Hey just wondering how is this moving along ? It'd be great if it were to solve windows machines getting stuck on bundling (although there is the workaround of using --debug).
Thanks for your work !

nddery:
Finally got to install a Windows VM and here's what I've noticed so far.

The result of `path.posix.relative(config.paths.ROOT, templatePath)` is wrong. Given that `config.paths.ROOT` is something like `C:\Users\nddery` on Windows, the above translate to something like `path.posix.relative('C:\Users\nddery', './src/templates/NotFound')`.

Running it in Node on Window, I do get strange results:

```node
> process.cwd()
'C:\\Users\\nddery\\react-static'

> path.posix.relative('C:\Users\nddery\react-static', './packages/react-static')
'../packages/react-static'
```

Note the extra `../` there. Wondering if mixing path style in `path.posix.relative` is tripping it up or something ? I've found https://github.com/nodejs/node/issues/17413 which mentions that " windows has a separate cwd for each drive", but not sure how this all applies here and how to address it.

nddery:
@SleeplessByte I've debugged a bit further and from what I can see, not using `path.posix` fixes the issue.

Seems like when using `path.posix.relative` the output is wrong, but while using `path.relative`, it is correct:

```node
> path.posix.relative('C:\\Users\\nddery\\react-static', './packages/react-static')
'../packages/react-static'

> path.relative('C:\\Users\\nddery\\react-static', './packages/react-static')
'code\\react-static'
```

Since we're using `slash` in `extractTemplates`, Windows backslash will be converted to slashes already.

This will result in links like `__react_static_root__/src/path/to/whatever` when `config.paths.dist` points inside the project root while it will end up with links like `__react_static_root__/../../tmp` when `config.path.dist` points outside of the project root. The `__react_static_root__` part ensures we don't add absolute paths to the generated bundles.

Can't seem to create a pull request that points to this pull request (as the branch for it is on your account), so [here's the commit where I fixed the tests for Windows](https://github.com/nddery/react-static/commit/a7cec429be796f2df3142f2e0c455b5baa811c9a).

Let me know what you think!

SleeplessByte:
I think that could be it. The paths have always been shady. I've tried to increase cross-env compatibility by using posix, but the fact that we're not using it _everywhere_ is probably, like you see emperically, what caused the bugs.

I'll merge in your commit and we'll try it out.

julesce:
Just a quick note to say thanks for your time and effort on this, it is much appreciated!

Issue #1021, opened 2019-02-15
[Bug] Lerna + react hooks


meesvandongen:
### Bug

When running using lerna & react hooks, react-static is not able to export. 
Instead, this error is shown: 
`Invariant Violation: Failed exporting HTML for URL / (src\pages\index.tsx): Minified React error #307; visit https://reactjs.org/docs/error-decoder.html?invariant=307`

According to that page, this is the error: `Hooks can only be called inside the body of a function component`

If I manually place the global `node_modules` in the local project folder, the build succeeds.

### Environment

```
  System:
    OS: Windows 10
    CPU: (8) x64 Intel(R) Core(TM) i7-2700K CPU @ 3.50GHz
    Memory: 6.84 GB / 15.95 GB
  Binaries:
    Node: 11.9.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.13.0 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 6.5.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: 44.17763.1.0
    Internet Explorer: 11.0.17763.1
  Dependencies:
    react: 16.8.1
    react-dom: 16.8.1
    react-static: 6.0.18
```

### Steps to Reproduce the problem

1.  `lerna init`
2.  configure lerna for yarn workspaces
3.  copy template to `/packages` & provide package version
4.  run `lerna bootstrap`
5.  run `yarn build`

### Expected Behavior

The build should succeed

### Reproducible Demo

https://github.com/meesvandongen/bug-lerna-reactstatic

tannerlinsley:
This is a bug with react in general when using lerna or linked dependencies. We're working on a workaround that will hopefully fix this.

meesvandongen:
I bypassed this by switching to pnpm instead of yarn+lerna for management.

IOIO72:
Is it really related to lerna? I'm getting a similar error without lerna.

Everything is fine in dev mode - as soon as I build the site, I'm getting the same error.

Here's the output of debug mode:

```
=> Exporting HTML...
  Error: Invariant Violation: Failed exporting HTML for URL 404 (src/pages/404.js): Hooks can only be called inside the body of a function component. (https://fb.me/react-invalid-hook-call)

  - react.development.js:88 invariant
    [honma]/[react]/cjs/react.development.js:88:15

  - react.development.js:1436 resolveDispatcher
[...]

PR #1383, opened 2020-02-17
Fix monorepo node externals resolution

### LINKED ISSUE #1216: Lerna / React Hooks isssue (unsure)
I am still experiencing this issue—has a solution been implemented yet?

I have a _very simple_ Lerna monorepo with 2 packages: `xrc` and `xrc-app`

`xrc` is the package I'm publishing to NPM and `xrc-app` is the docs website/app for `xrc`

`xrc-app` is _private_ and uses `react-static`

https://github.com/wagerfield/xrc

I have purposefully removed `react` and `react-dom` as dependencies from both `xrc` and `xrc-app` to be absolutely sure that the issue I'm experiencing isn't related to multiple versions of `react` being installed.

After running `yarn` from the root of the monorepo to install the dependencies, I then run `yarn why react` and get:

```
yarn why v1.16.0
[1/4] 🤔  Why do we have the module "react"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "react@16.8.6"
info Reasons this module exists
   - "_project_#xrc-app#react-static" depends on it
   - Hoisted from "_project_#xrc-app#react-static#react"
info Disk size without dependencies: "248KB"
info Disk size with unique dependencies: "564KB"
info Disk size with transitive dependencies: "648KB"
info Number of shared dependencies: 6
```

This clearly shows that I only have _one_ version of `react` installed (16.8.6) which is the result of `react-static` listing it as a dependency.

Since the `xrc-app` lives inside the `packages` dir, it uses the local version of `xrc` rather than the one published to NPM. Therefore, before you can build the React Static site, you _must run_ `yarn bundle` from the root of the monorepo. This will compile the TypeScript in the `xrc` package using Rollup into `cjs` and `esm` bundles in the `xrc` package that `xrc-app` can then import and use.

After running `yarn bundle` you can now run all of the `app:*` related scripts from the monorepo root.

When you run `yarn app:start` from the monorepo root, it will run the `start` script in the `xrc-app` package which in turn calls `react-static start` to start the development server. This works absolutely fine and you will be able to see the WIP site at `localhost:5000`

However, when you run `yarn app:stage` or `yarn app:build` from the monorepo root it will get as far as compiling all the JS bundles, but then it hangs and eventually fails when exporting the HTML.

I then get the error:

```
Invariant Violation: Failed exporting HTML Minified React error #321
```

https://reactjs.org/docs/error-decoder.html?invariant=321

1. You might have mismatching versions of React and the renderer (such as React DOM)
2. You might be breaking the Rules of Hooks
3. You might have more than one copy of React in the same app

See https://fb.me/react-invalid-hook-call for tips about how to debug and fix this problem.

As mentioned above, running `yarn why react` and `yarn why react-dom` from the monorepo root will show that only one version of both `react` and `react-dom` are installed and that they are the _same version_ of `16.8.6`.

Therefore I would assume that I'm not falling foul of point 1 or 3. Furthermore, I'm not using hooks anywhere in my application code.

If it is an issue with React hooks, then perhaps it's coming from React Static?

The strange thing I have found is that if I copy and paste the `xrc-app` package onto my desktop so that it's completely removed from the Lerna monorepo and run `yarn` or `npm install` inside the package so `react-static` et al are installed and `xrc` is pulled from NPM rather than the local monorepo version, then the `build` and `stage` scripts work.

I've been battling with this one all day and could really use some help please :D

Thanks in advance!

_Originally posted by @wagerfield in https://github.com/nozzle/react-static/issues/1021#issuecomment-499607184_

kitten:
Fix #1216 

## Description

This replaces `webpack-node-externals` with a `resolve-from` based `externals` config in `webpack.config.prod.js` that fixes the monorepo support in `react-static`.

## Changes/Tasks

- Rewrite `externals` Webpack option for `node` stage in `webpack.config.prod.js`
- Remove `webpack-node-externals` package

## Motivation and Context

`webpack-node-externals` is not properly maintained and doesn't take monorepos and projects with multiple `node_modules` folders into account. Realistically we'd want to be as close to the node resolution algorithm as possible. Hence this PR utilises `resolve-from`.

This change is similar to the logic that is present in `next.js` for the same purpose: https://github.com/Timer/next.js/blob/31a47b4bffa45fc9d5f2bd464660bb68f6037229/packages/next/build/webpack-config.ts#L490-L548

## Types of changes

## Checklist:

### DIFF
--- a/packages/react-static/src/static/webpack/webpack.config.prod.js
+++ b/packages/react-static/src/static/webpack/webpack.config.prod.js
@@ -3,10 +3,9 @@ import path from 'path'
 import CaseSensitivePathsPlugin from 'case-sensitive-paths-webpack-plugin'
 import { BundleAnalyzerPlugin } from 'webpack-bundle-analyzer'
 import TerserPlugin from 'terser-webpack-plugin'
-import nodeExternals from 'webpack-node-externals'
 import ExtractCssChunks from 'extract-css-chunks-webpack-plugin'
 import OptimizeCSSAssetsPlugin from 'optimize-css-assets-webpack-plugin'
-import resolveFrom from 'resolve-from'
+import { silent as resolveFrom } from 'resolve-from'
 //
 import rules from './rules'
 
@@ -165,18 +164,29 @@ export default function(state) {
   result.devtool = false
   result.externals = [
     new RegExp(`${paths.PLUGINS}`),
+    // Externalise all node_modules except for react-static files that need to be bundled
     (context, request, callback) => {
-      const resolved = path.resolve(context, request)
-      if (
-        [/react-static(\\|\/)lib(\\|\/)browser/, /webpack-flush-chunks/].some(
-          d => d.test(resolved)
-        )
-      ) {
-        return callback(null, `commonjs ${resolved}`)
+      // All paths starting with `./` or `/` may be bundled as usual
+      if (/^[./\\]/.test(request)) {
+        return callback();
       }
-      callback()
+
+      // Attempt to resolve the requested module or file from `context` which points at
+      // the file's location where the dependency on `request` is located at.
+      const res = resolveFrom(`${context}/`, request);
+
+      if (
+        res &&
+        // If the request is in node_modules and a js file it should be externalised,
+        /node_modules[/\\].*\.js$/.test(res) &&
+        // unless it's a react-static dependency
+        !/node_modules[/\\]react-static/.test(res)
+      ) {
+        callback(null, `commonjs ${request}`);
+      } else {
+        callback();
+      }
     },
-    nodeExternals(),
   ]
   result.module.rules = rules(state)
   result.plugins = [
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 - Fix stderr pollution by progress module ([#1356](https://github.com/react-static/react-static/pull/1356))
 - Fix package.json and README for `react-static-plugin-stylus` ([#1244](https://github.com/react-static/react-static/issues/1244))
 - Fix Webpack stats output in environment that don't support color ([#1370](https://github.com/react-static/react-static/pull/1370))
+- Fix Webpack `externals` configuration for monorepo support ([#1383](https://github.com/react-static/react-static/pull/1383))
 
 ## 7.2.2
 
--- a/packages/react-static/package.json
+++ b/packages/react-static/package.json
@@ -102,8 +102,7 @@
     "webpack": "^4.39.2",
     "webpack-bundle-analyzer": "^3.4.1",
     "webpack-dev-server": "^3.8.0",
-    "webpack-flush-chunks": "^2.0.3",
-    "webpack-node-externals": "^1.7.2"
+    "webpack-flush-chunks": "^2.0.3"
   },
   "devDependencies": {
     "@types/react": "^16.9.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -796,6 +796,13 @@
   dependencies:
     regenerator-runtime "^0.13.2"
 
+"@babel/runtime@^7.3.1", "@babel/runtime@^7.6.2":
+  version "7.8.4"
+  resolved "https://registry.yarnpkg.com/@babel/runtime/-/runtime-7.8.4.tgz#d79f5a2040f7caa24d53e563aad49cbc05581308"
+  integrity sha512-neAp3zt80trRVBI1x0azq6c57aNBqYZH8KhMm3TaB7wEI5Q4A2SHfBHE8w9gOhI/lrqxtEbXZgQIrHP+wvSGwQ==
+  dependencies:
+    regenerator-runtime "^0.13.2"
+
 "@babel/template@^7.1.0", "@babel/template@^7.4.0", "@babel/template@^7.4.4":
   version "7.4.4"
   resolved "https://registry.yarnpkg.com/@babel/template/-/template-7.4.4.tgz#f4b88d1225689a08f5bc3a17483545be9e4ed237"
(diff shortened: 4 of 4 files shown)


packages/react-static/src/static/webpack/webpack.config.prod.js
```diff
@@ -3,10 +3,9 @@ import path from 'path'
 import CaseSensitivePathsPlugin from 'case-sensitive-paths-webpack-plugin'
 import { BundleAnalyzerPlugin } from 'webpack-bundle-analyzer'
 import TerserPlugin from 'terser-webpack-plugin'
-import nodeExternals from 'webpack-node-externals'
 import ExtractCssChunks from 'extract-css-chunks-webpack-plugin'
 import OptimizeCSSAssetsPlugin from 'optimize-css-assets-webpack-plugin'
-import resolveFrom from 'resolve-from'
+import { silent as resolveFrom } from 'resolve-from'
```
kitten: I assume this change is safe because we can assume that the listed alises in `resolve.alias` that use this utility, namely `react`, `react-dom`, and `react-universal-component`, are all assumed to be installed

SleeplessByte (approved):
It looks good so far.

Can you explain how we could test this? Set up a monorepo with _which_ folders using react-static? How will it fail before this fix and how does it not-fail after?

kitten:
@SleeplessByte Realistically it’s hard to test in-repo unless you have examples folders set up and are willing to drop in a full monorepo structure into one of these folders? I’d have to defer for a decision to you on that.

I’ve noticed however that you don’t change the Webpack config very often. So maybe a local test would suffice?

To get to what this actually fixes, suppose you have a repository that is set up with Lerna or Yarn Workspaces with a root `package.json` and a package with `packages/site/package.json`. The `site` in this example would contain and build a `react-static` site.

Typically and by default all dependencies would be hoisted to the root `node_modules` folder including `react` and `react-dom`. Without this fix the `externals` config with `webpack-node-externals` would cause all packages in the **local** `node_modules` folder not to be bundled by telling Webpack to require them using the `commonjs` prefix. In a monorepo scenario with hoisting most dependencies will however be in the root `node_modules` folder.

This causes dependencies to be accidentally bundled in the Node bundle. This isn’t desired and the first sign that this has happened is that we’ll see duplicate React errors due to the React Dispatcher being duplicated.

After this fix all resolvable packages in all `node_modules` folders won’t be bundled which fixes this issue.

Hope that clarifies this PR 👍 Let me know if you’d like me to set up an example repo or what you’d like to do about testing

SleeplessByte:
A local test suffices 🗡 . We absolutely don't need this as _test_ inside this repo 👍 

I'll try out your steps. Thank you very much for the explanation. It's very clear :)

bram-l:
I also ran into this same issue #1216 and #1021 using Lerna & Yarn Workspaces and I can confirm that after applying this patch everything worked as expected!

Just out of curiosity, why was the exception for `/react-static(\\|\/)lib(\\|\/)browser/, /webpack-flush-chunks/` removed?

mikestopcontinues:
Works like a charm for my lerna/npm monorepo needs!

packages/react-static/src/static/webpack/webpack.config.prod.js
```diff
@@ -165,18 +164,29 @@ export default function(state) {
...
-      const resolved = path.resolve(context, request)
+      // All paths starting with `./` or `/` may be bundled as usual
+      if (/^[./\\]/.test(request)) {
+        return callback();
+      }
+
+      // Attempt to resolve the requested module or file from `context` which points at
+      // the file's location where the dependency on `request` is located at.
+      const res = resolveFrom(`${context}/`, request);
+
```
SleeplessByte, line 177: Shouldn't this also include this loader check/webpack check: https://github.com/Timer/next.js/blob/31a47b4bffa45fc9d5f2bd464660bb68f6037229/packages/next/build/webpack-config.ts#L536-L542

packages/react-static/src/static/webpack/webpack.config.prod.js
```diff
@@ -165,18 +164,29 @@ export default function(state) {
...
       if (
-        [/react-static(\\|\/)lib(\\|\/)browser/, /webpack-flush-chunks/].some(
-          d => d.test(resolved)
-        )
+        res &&
+        // If the request is in node_modules and a js file it should be externalised,
+        /node_modules[/\\].*\.js$/.test(res) &&
+        // unless it's a react-static dependency
+        !/node_modules[/\\]react-static/.test(res)
```
nddery, line 183: After https://github.com/react-static/react-static/pull/1444, we'd need to update this like so:

```suggestion
        !/node_modules[/\\]react-static/.test(res) &&
        // react-universal-component must be bundled for static export to work
        // https://github.com/react-static/react-static/issues/1446
        !/node_modules[/\\]react-universal-component/.test(res)
```

Part of me think it might be cleaner to put it in its own `if` statement ?

SleeplessByte: Yeah I think we just need th entire list next.js is doing too

ar7casper:
Hi, 

Do you have an ETA on merging this PR?
I'm using a monorepo and I can't make it work.
If I understand correctly, #1493 is related to this PR?

packages/react-static/src/static/webpack/webpack.config.prod.js
```diff
@@ -165,18 +164,29 @@ export default function(state) {
...
     (context, request, callback) => {
-      const resolved = path.resolve(context, request)
+      // All paths starting with `./` or `/` may be bundled as usual
+      if (/^[./\\]/.test(request)) {
+        return callback();
+      }
+
+      // Attempt to resolve the requested module or file from `context` which points at
+      // the file's location where the dependency on `request` is located at.
+      const res = resolveFrom(`${context}/`, request);
```
merceyz, line 176: You can use `require.resolve` for this so you don't need a external dependency https://nodejs.org/api/modules.html#modules_require_resolve_request_options

SleeplessByte:
@ar7casper Various unresolved comments on this PR; so it can't be merged at the moment.

DBosley:
This PR has been open for a long time. Is there a chance these last comments can get resolved so it can be merged?

SleeplessByte:
@DBosley only if someone resolves the open comments in this PR.

PR #1389, opened 2020-02-27
Fix registration of 404 route info

bram-l:
Make sure a 404 template is registered for the '404' path when prefetching (instead of the original path). This will prevent tv non-existing route data for a 404 page.

## Description

This was previously for V6, but apparently got lost during the refactoring of V7. See: #1014  

## Types of changes

### DIFF
--- a/packages/react-static/src/browser/components/Routes.js
+++ b/packages/react-static/src/browser/components/Routes.js
@@ -9,7 +9,9 @@ import {
   prefetch,
   plugins,
   onReloadTemplates,
+  routeErrorByPath,
 } from '..'
+import { getCurrentRoutePath } from '../utils'
 import { useStaticInfo } from '../hooks/useStaticInfo'
 import { routePathContext, useRoutePath } from '../hooks/useRoutePath'
 
@@ -74,6 +76,16 @@ const RoutesInner = ({ routePath, render: renderFn }) => {
       // In SRR and production, synchronously register the template for the
       // initial path
       registerTemplateForPath(path, template)
+
+      // For a 404 route we will register the current route as invalid
+      if (path === '404') {
+        const currentPath = getCurrentRoutePath()
+        // As long as we didn't navigate to the 404.html page directly
+        if (currentPath !== '404') {
+          routeErrorByPath[currentPath] = true
+          templateErrorByPath[currentPath] = true
+        }
+      }
     }
   })
 
@@ -88,7 +100,7 @@ const RoutesInner = ({ routePath, render: renderFn }) => {
     routePath = staticInfo.path
   } else if (!routePath) {
     // If a routePath is still not defined in the browser,
-    // use the window location as the defualt
+    // use the window location as the default
     routePath = decodeURIComponent(window.location.href)
   }
 
--- a/packages/react-static/src/browser/components/Routes.js
+++ b/packages/react-static/src/browser/components/Routes.js
@@ -11,7 +11,7 @@ import {
   onReloadTemplates,
   routeErrorByPath,
 } from '..'
-import { getCurrentRoutePath } from '../utils'
+import { getCurrentRoutePath, is404Path, PATH_404 } from '../utils'
 import { useStaticInfo } from '../hooks/useStaticInfo'
 import { routePathContext, useRoutePath } from '../hooks/useRoutePath'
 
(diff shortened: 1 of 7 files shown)


bram-l:
@SleeplessByte could you have a look at this?

packages/react-static/src/browser/components/Routes.js
```diff
@@ -74,6 +76,16 @@ const RoutesInner = ({ routePath, render: renderFn }) => {
       // In SRR and production, synchronously register the template for the
       // initial path
       registerTemplateForPath(path, template)
+
+      // For a 404 route we will register the current route as invalid
+      if (path === '404') {
```
SleeplessByte: I think we allow other paths to be 404 as well. I don't recall if there is a function to give us the 404 path, but if there is, that should be used instead (for both `path` and `currenPath`.

Other than that, this looks great, except that we're not testing it. Would that be hard?

bram-l: You mean to define a constant for the 404 path? Or a helper function for checking whether a path is the '404' path? Now that I'm thinking about it, maybe both woud be best.. But, since this value is used in browser specific logic (eg `src/browser/index.js`)  as well as for the static site generation (eg. `src/static/getRoutes.js`), what would a suitable place for such constant / helper? 

Regarding the testing, I think this will be tricky. The goal is to prevent a 404 page entering in a loading state while the required route data is already embedded in the HTML. Unfortunately a simple unit won't cover this, so it would require a full end-to-end test, but I don't think the current Travis setup supports this right?

SleeplessByte: > but I don't think the current Travis setup supports this right?

Yeah the current setup makes that very hard. The next major should have it out of the box, but with this COVID-19, our hands are... elsewhere, so it's taking its time.

> You mean to define a constant for the 404 path

I mean that we already output `is404` sometimes (it's a boolean). I believe that is determined on more than checking if the path is /404. I would use _that_ logic, if it's different. Does that make sense?

bram-l: > I believe that is determined on more than checking if the path is /404. I would use that logic, if it's different. 

The `is404` comes from [this line](https://github.com/react-static/react-static/blob/master/packages/react-static/src/browser/components/Routes.js#L22), and indeed it's just checking the path. One thing to note though is that it's not checking the path on the current browser location, but rather the route info. For example a non-existing page `/foo` would be served the `/404.html`. So when we you land on that page the `path` in the `staticInfo` will be `"404"`. In that case to make sure that it won't try to fetch `/foo/routeData.json`, we need to register the 404 template for the `/foo` path, which is done [here](https://github.com/react-static/react-static/blob/89c02a16f988b5f352da4890b4b89ed40f77e977/packages/react-static/src/browser/components/Routes.js#L82-L86).

So to come back to the question of abstracting the `is404` logic, I think what we could do is define a constant for 404 instead of relying on the string literal. But I think that only makes sense if the same constant is being used during the static site generation, for example [here](https://github.com/react-static/react-static/blob/master/packages/react-static/src/static/getRoutes.js#L120) and [here](https://github.com/react-static/react-static/blob/master/packages/react-static/src/static/getRoutes.js#L48). Right?

SleeplessByte: Yes! Exactly right :)

bram-l: I've added the helper function as well as the 404 path constant. I hope this was more or less this what you had in mind. 

Regarding the testing, I've tried to put together a simple unit test but that turned out to be rather challenging. The `Routes` component throws a Promise which will need to be caught by a `Suspense` component, but it seems like Enzyme doesn't support Suspense yet, at least not when the fallback is triggered. I don't see how I could test this without changing the entire test setup.. any suggestions?

SleeplessByte: > The Routes component throws a Promise which will need to be caught by a Suspense component.

No if the data is ready / present! But it's a bit difficult to set up right now.

Does it work with your local project(s)? If so, I'll check out your branch and try it on ours :)

bram-l: Yes we actually run this in production already using our own fork, all working smoothly!

Issue #1093, opened 2019-04-03
[Bug] reloadClientData() does not reload site data


Blobson:
**Short description**
I am using **site.yaml** file to store site data and in development mode I want to hot-reload site data on every change in **site.yaml**.

**Steps to reproduce the issue**
1. Create new project using **react-static v7.0.5** using blank template.
2. Add **data/site.yaml** with following contents to the project:
```yaml
title: My Site!
```
3. You will need to install **node-watch** and **js-yaml** for the next step:
```
$ yarn add node-watch js-yaml
```
4. Add **getSiteData** and a watcher for **data/site.yaml** inside **static.config.js** like this:
```javascript
import fs from 'fs'
import path from 'path'
import yaml from 'js-yaml'
import watch from 'node-watch'
import { reloadClientData } from 'react-static/node'

const dataDir = path.join(__dirname, 'data')

watch(dataDir, { recursive: true, filter: /\.(?:yaml|yml)$/ }, (evt, name) => {
  console.log(`Detected changes in '${name}',  reloading client data...`)
  reloadClientData()
})

export default {
  maxThreads: 1, // Remove this when you start doing any static generation
  getSiteData: () => {
    try {
      const rawSiteData = fs.readFileSync(path.join(dataDir, 'site.yaml'))
      const yamlSiteData = yaml.safeLoad(rawSiteData)
      console.log('SiteData: ', yamlSiteData)
      return yamlSiteData
    } catch(e) {
      console.error(e)
      return {}
    }
  }
}
```
5. Add use of `const { title } = useSiteData()` to the **src/App.js** component like this:
```javascript
import React, { Suspense } from 'react'
import { useSiteData } from 'react-static'

import './app.css'
import logo from './logo.png'

const Title = () => {
  const { title } = useSiteData()
  return (
    <h1>{title}</h1>
  )
}

const App = () => (
  <Suspense fallback="Loading...">
    <div className="App">
      <header className="App-header">
        <img src={logo} className="App-logo" alt="logo" />
        <Title />
      </header>
    </div>
  </Suspense>
)

export default App
```
6. Start development server
7. Try editing **title** in **data/site.yaml** and you'll see that changes does not hot-reload.
Also in the stdout of dev server you'll see that `reloadClientData()` is called but it does not trigger call for `getSiteData()`.

**Solution that worked for me**
For now the only solution that worked for me is to add: 
```javascript
latestState = await fetchSiteData(latestState)
``` 
[...]

PR #1409, opened 2020-04-24
Add call to fetchSiteData when reloadClientData is called in runDevserver.js

krishan711:
## Description

Fix for https://github.com/react-static/react-static/issues/1093

The suggestion in the issue above was to update the latestState variable directly, but as mentioned this doesn't work. I believe this is primarily because the routes are set already and `before` is not called when the webpack dev server is invalidated. 

Testing locally this works, just wanted a second opinion before i look into adding tests as it may be overly heavy-handed.

## Changes/Tasks

- [ ] Updated to call buildDevRoutes
- [ ] (TODO): add tests if this is agreed to be a viable solution

## Motivation and Context

comes from https://github.com/react-static/react-static/issues/1093. In summary, if you try to call `reloadClientData` in static.config.js the reload happens but the data isn't actually re-retrieved.

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 - Fix stderr pollution by progress module ([#1356](https://github.com/react-static/react-static/pull/1356))
 - Fix package.json and README for `react-static-plugin-stylus` ([#1244](https://github.com/react-static/react-static/issues/1244))
 - Fix Webpack stats output in environment that don't support color ([#1370](https://github.com/react-static/react-static/pull/1370))
+- Fix reloadClientData to call fetchSiteData runDevserver.js ([#1409](https://github.com/react-static/react-static/pull/1409))
 
 ## 7.2.2
 
--- a/packages/react-static/src/static/webpack/runDevServer.js
+++ b/packages/react-static/src/static/webpack/runDevServer.js
@@ -134,7 +134,7 @@ async function runExpressServer(state) {
                 if (!route) {
                   const err = new Error(
                     `Route could not be found for: ${routePath}
-                    
+
 If you removed this route, disregard this error.
 If this is a dynamic route, consider adding it to the prefetchExcludes list:
 
@@ -226,6 +226,7 @@ If this is a dynamic route, consider adding it to the prefetchExcludes list:
   const socket = io()
 
   reloadClientData.current = async () => {
+    await buildDevRoutes(latestState);
     socket.emit('message', { type: 'reloadClientData' })
   }
 
--- a/packages/react-static/src/static/webpack/runDevServer.js
+++ b/packages/react-static/src/static/webpack/runDevServer.js
@@ -14,6 +14,7 @@ let devServer
 let latestState
 let buildSiteDataRoute = () => {}
 let buildDevRoutes = () => {}
+const getLatestState = () => latestState
 
 export const reloadClientData = () => {
   if (reloadClientData.current) {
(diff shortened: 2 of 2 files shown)


SleeplessByte:
We can't just merge this; it definitely needs a flag. `routes` can take _minutes_ to build depending on your setup. You don't want a  Fast Refresh / Hot Reload to take minutes. 

Instead, what the author of #1093 wants is that the `siteData` is cleared (and reloaded), instead of _just_ the route data. The route data is composed of the routes.data + site.data.

krishan711:
@SleeplessByte Sure, makes sense. How would you suggest going about it since the current code doesn't refresh the site data? Would it make sense to split out a function that builds the site data route (`'/__react-static__/siteData'`) so i can call that function individually and not affect the other routes?

SleeplessByte:
That would not be a bad idea. How would you go about to merge that data back into the routeData? As in, what's the idea there to update the routeData as the siteData changes?

krishan711:
im not totally sure what you mean here, probably because I haven't used routeData and siteData together. why would routeData need to be reloaded if the siteData changes? or where should I look to understand this? I checked https://github.com/react-static/react-static/blob/master/packages/react-static/src/static/getRouteData.js but there is no reference to the siteData. is it because the context has state passed to `getData` has the siteData in it and people may have code dependent on the siteData?

krishan711:
actually maybe i dont understand the original comment anymore - it doesnt look like `routes` is rebuilt in this function from what i can see? it's built in start.js and then passed in via the state and not ever updated from what i can tell.

SleeplessByte:
Sorry, I was confusing you 😓 😅 . I wasn't talking about `siteData` but `sharedData`, as you can see here: https://github.com/react-static/react-static/blob/611aa824e15f19466911623705dc0676dbe9c388/packages/react-static/src/browser/utils/index.js#L230-L235

Anyway. 

> @SleeplessByte Sure, makes sense. How would you suggest going about it since the current code doesn't refresh the site data? Would it make sense to split out a function that builds the site data route ('/__react-static__/siteData') so i can call that function individually and not affect the other routes?

Yes, this works perfectly. I just checked and that's how we have it set up for some of our local forks!

krishan711:
Ah awesome, I'll try this out this evening and post an update then. Thanks 👍

krishan711:
Hm, I updated this but actually this doesn't seem to work anymore and I can't see immediately why. Will dig deeper.

krishan711:
Okay, im no longer sure why things weren't working before. This seems to work for me now, contrary to what the open issue says. @SleeplessByte is this good or am i missing something here?

krishan711:
Hm, in the issue they only tried to change the code directly in the node_modules folder. This didnt work for me. When i edited the source and published to my own registry and used that instead of react-static everything works as expected. You can try it yourself with "@kibalabs/react-static": "7.2.3-kiba.2"`

SleeplessByte:
Hah. You mean it's just working, always :P ?

I don't have the capacity right now to try it out, but we can always close this and reopen if you find a circumstance where it doesn't work.

krishan711:
Oh no sorry I meant it works with the addition of the single line i have added in the PR now

krishan711:
(it is what was suggested in the open issue)

SleeplessByte:
It looks fine for me. Tests still pass. I doesn't look breaking. 

Can you add this to CHANGELOG.md (same format, bottom of the `master` list)? We'll merge it for 7.4.0 and if it creates bugs we can revert or revisit.

krishan711:
done. thanks!

SleeplessByte:
Thank you for the effort @krishan711

Issue #1361, opened 2020-01-11
[Bug] Internet explorer does not load routes


BrendonSled:
### Bug

Internet explorer with compatibility found in the guide does not work. Static routes will show a 404 and when using things like `useRouteData` or `withRouteDate` the site will never load. There are also no console.error messages so I'm not sure where the issues lies.

### Environment

```

  System:
    OS: Windows 10 10.0.18363
    CPU: (12) x64 AMD Ryzen 5 3600 6-Core Processor
    Memory: 6.76 GB / 15.93 GB
  Binaries:
    Node: 12.13.0 - C:\Program Files\nodejs\node.EXE
    npm: 6.12.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: 44.18362.449.0
  npmPackages:
    react: ^16.9.0 => 16.12.0
    react-dom: ^16.9.0 => 16.12.0
    react-hot-loader: ^4.12.11 => 4.12.18
    react-static: ^7.2.0 => 7.2.3
    react-static-plugin-reach-router: ^7.2.0 => 7.2.3
    react-static-plugin-sitemap: ^7.2.0 => 7.2.3
    react-static-plugin-source-filesystem: ^7.2.0 => 7.2.3
    react-static-plugin-typescript: ^7.2.0 => 7.2.3
    babel-polyfill: "^6.26.0",

```

### Steps to Reproduce the problem

**Base your steps off of any freshly installed react-static template!**

1.  Use  react-static template typescript (I'm sure others work)
2.  Add IE compatibility
```
    webpack: (config, { stage }) => {
        config.entry = [
            'babel-polyfill',
            ...(Array.isArray(config.entry) ? config.entry : [config.entry]),
        ];
        return config;
    },
```
3.  `npm start`

### Expected Behavior

Routes to load fine

SleeplessByte:
Which IE version exactly?

BrendonSled:
IE 11

SleeplessByte:
Hmm. I don't know. I know we don't really support IE at all (because it takes too much time to give support for), but it could well be that we've inadvertently broken this. If you figure out _why_ it's failing, that's great.

The first place I would look at is check if `Promise` is actually polyfilled, because those API's use promises.

BrendonSled:
Everything is polyfilled. I've made a sort of work around for now.
```jsx
<Routes
    render={({ routePath, getComponentForPath }) => {
        // The routePath is used to retrieve the component for that path
[...]

PR #1461, opened 2020-06-23
Updated browser support docs for IE 11

BrendonSled:
These changes bring support for older browsers with updated polyfills

## Motivation and Context

Over releases browser support for IE 11 has stopped. This PR brings compatibility. 
Fixes issues #1361

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,12 +4,12 @@
 
 ### New
 
-- Updated browser support docs for IE11 support
-
 ### Improved
 
 ### Bugfix
+
 - Ensure a non-zero exit status code when build has errors ([#1451](https://github.com/react-static/react-static/pull/1451))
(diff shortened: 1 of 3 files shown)


Sut3kh:
Nice, avoiding the silent error should hopefully save someone a lot of time one day!

Please correct me if I am wrong but hasn't `webpack` been removed from static.config.js to encourage the use of the plugin system for this? 

I didn't really look to far into it at the time but that's what we had to do to modify webpack config (in 7.2.3)

SleeplessByte:
I don't think webpack has been removed @Sut3kh -- but regardless the changes to be made are the same :) (but then as plugin). You are correct that we favour plugin over webpack direct. In RS 8, this will be easier, and a webpack hook is available still.

Issue #1331, opened 2019-10-24
[Bug] Preact plugin seems to not work


jurosh:
### Reporting a bug?

Trying to use preact plugin - https://github.com/react-static/react-static/tree/master/packages/react-static-plugin-preact

But even after including `preact` in plugins, there is no difference to bundle size. Still `analyze` display around 400kB for `react` & `react-dom`, and it should be around 3kB for `preact`.

![image](https://user-images.githubusercontent.com/1630137/67478604-a8260980-f65c-11e9-8f47-d6ad294260c7.png)

### Environment

```
System:
    OS: Ubuntu 19.10
  Binaries:
    Node: 11.14.0 - ~/.nvm/versions/node/v11.14.0/bin/node
    Yarn: 1.19.1 - /usr/bin/yarn
    npm: 6.11.3 - ~/.nvm/versions/node/v11.14.0/bin/npm
  npmPackages:
    react-static: ^7.2.2 => 7.2.2 
    react-static-plugin-preact: ^7.2.2 => 7.2.2 
    react-static-plugin-reach-router: ^7.2.0 => 7.2.2 
    react-static-plugin-sitemap: ^7.2.0 => 7.2.2 
    react-static-plugin-source-filesystem: ^7.2.0 => 7.2.2
```

### Steps to Reproduce the problem

**Base your steps off of any freshly installed react-static template!**

1.  add preact plugin
2.  run `yarn analyze`
3.  see that react is there

SleeplessByte:
It currently only applies preact in production, or so it seems:

https://github.com/react-static/react-static/blob/master/packages/react-static-plugin-preact/src/node.api.js#L3

SleeplessByte:
If you'd like this to work in dev too, we accept a PR. It won't be a bug but a new feature :)

tbgse:
@SleeplessByte I know this issue has long been closed but i'm seeing the same problem even after creating a custom plugin with the same preact replacement code without the `dev` condition:

Also following the original issue above, wouldn't `yarn analyze` create a production bundle where the original plugin should've been applied? Very curious if @jurosh ever found a solution to it.

here is my plugin code:

```
export default () => ({
  webpack: (config, { stage }) => {
    config.resolve.alias = {
      ...(config.resolve.alias || {}),
      react: 'preact-compat',
      'react-dom': 'preact-compat',
    }
    
    return config
  },
})
```

in `plugins/custom-preact/node.api.js`

[...]

PR #1486, opened 2020-08-07
Update preact plugin

tbgse:
## Description
The preact plugin is currently not functioning due to the alias pointing to the wrong references. See #1331

## Changes/Tasks
- Upgrade preact to latest version
- Remove preact-compat (is now included in preact core)
- Change alias to correctly reference the actual react packages
- Add prefresh to replace react-hot-reload

## Motivation and Context

see #1331 

## Types of changes

## Checklist:

Additional Notes: I'm not 100% sure how the dependency management works with yarn workspaces and also unsure how to test the actual code on this branch with an example project. Please let me know what I can do to pull in the plugin from this branch into a fresh react-static project.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -250,9 +250,9 @@ rebuildRoutes()
 
 // Reload when files change
 import chokidar from 'chokidar'
-// Ensure routes are built at least once before rebuilding routes 
+// Ensure routes are built at least once before rebuilding routes
 let areRoutesBuilt = false
-chokidar.watch('./docs').on('all', () => isReady  && rebuildRoutes())
+chokidar.watch('./docs').on('all', () => isReady && rebuildRoutes())
 
 // Reload from API or CMS event
 YourFavoriteCMS.subscribe(rebuildRoutes)
--- a/packages/react-static-plugin-preact/README.md
+++ b/packages/react-static-plugin-preact/README.md
@@ -17,3 +17,27 @@ export default {
   plugins: ["react-static-plugin-preact"]
 };
 ```
+
+As final step, depending on the template you started with, you might need to remove the React Hot Reload component from `index.js`. Change:
+
+```
+  const render = Comp => {
+    renderMethod(
+      <AppContainer>
+        <Comp />
+      </AppContainer>,
+      target
+    )
+  }
+```
+
+to:
+
+```
+  const render = Comp => {
+    renderMethod(
+      <Comp />,
+      target
+    )
+  }
+```
--- a/packages/react-static-plugin-preact/README.md
+++ b/packages/react-static-plugin-preact/README.md
@@ -41,3 +41,25 @@ to:
     )
   }
 ```
+
+To preserve component state when using prefresh, it is neccesary to name the components you're exporting. Instead of
+```
+export default () => {
+  return <p>Want to refresh</p>
+}
+```
+You'll have to write your components like this:
+```
+const Refresh = () => {
+  return <p>Want to refresh</p>
+}
+
+export default Refresh;
+```
+
+or
+```
+export default function Refresh () {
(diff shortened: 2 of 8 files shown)


docs/api.md
```diff
@@ -250,9 +250,9 @@ rebuildRoutes()
 
 // Reload when files change
 import chokidar from 'chokidar'
-// Ensure routes are built at least once before rebuilding routes 
+// Ensure routes are built at least once before rebuilding routes
 let areRoutesBuilt = false
-chokidar.watch('./docs').on('all', () => isReady  && rebuildRoutes())
+chokidar.watch('./docs').on('all', () => isReady && rebuildRoutes())
```
tbgse, line 256: seems like some linting was not done here before? This was applied after running `yarn test`

packages/react-static-plugin-preact/README.md
```diff
@@ -16,4 +16,28 @@ Then add the plugin to your `static.config.js`:
...
+to:
+
+```
+  const render = Comp => {
+    renderMethod(
+      <Comp />,
+      target
+    )
+  }
+```
```
tbgse, line 43: is there a way to apply this in a more automated way when the plugin is installed?

packages/react-static-plugin-preact/package.json
```diff
@@ -15,12 +15,11 @@
...
   },
   "peerDependencies": {
-    "preact": "^8.5.1",
-    "preact-compat": "^3.19.0"
+    "preact": "^10.4.7"
   },
   "dependencies": {
-    "preact": "^8.5.1",
-    "preact-compat": "3.19.0"
+    "@prefresh/webpack": "^0.8.2",
```
tbgse, line 21: eslint was complaining when i tried to install this as a dev dependency for the plugin and imported it in `node.api.js`. Should this be a regular dependency or dev dependency?

SleeplessByte: IIRC, `node.api` only runs when _compiling_ a react static site. To me that's a _development_ thing, so a devdependency. Do you per chance know which rule it was? It should be turned off for `node.api` files.

packages/react-static-plugin-preact/README.md
SleeplessByte: Yes, we could provide a `codemod`, or run a postinstall script.

docs/api.md
SleeplessByte: Prettier was added after this was written originally.

SleeplessByte:
I think this looks good for a first iteration. The suggestions can be applied in future PRs. Can you add this to CHANGELOG?

packages/react-static-plugin-preact/package.json
tbgse: the rule is `eslint(import/no-extraneous-dependencies)`

I just looked at the eslintignore and it seems like using only one star as wildcard is not enough to reach the nested src folder in the packages:

```
# Packages build into their root
packages/*/node.api.js
packages/*/browser.api.js
```

changing this to 
```
# Packages build into their root
packages/**/node.api.js
packages/**/browser.api.js
```
seems to correctly ignores the files

tbgse:
> I think this looks good for a first iteration. The suggestions can be applied in future PRs. Can you add this to CHANGELOG?

done, let me know if that fits the style you're going for with the changelog.

tbgse:
@SleeplessByte any other changes required to move this forward?

SleeplessByte (approved):
This is great. (Sorry. I get a few thousand notifications a week, so sometimes these slip through.)

SleeplessByte:
Thank you for your contribution @tbgse. It was [just released](https://github.com/react-static/react-static/releases/tag/v7.5.0).

PR #1473, opened 2020-07-11
fix: expand styled-component peer dep range

Flydiverny:
## Description

Expands version range in peer dependency of styled-components plugin
Not adding styled-components as a dependency as suggested in #1455 following recommendations from styled-components:
https://styled-components.com/docs/faqs#i-am-a-library-author-should-i-bundle-styledcomponents-with-my-library

## Changes/Tasks

- [ ] Changed code

## Motivation and Context

Resolves #1455 

## Types of changes

- [ ] Refactoring/add tests (refactoring or adding test which isn't a fix or add a feature)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x] I have updated the documentation accordingly
- [x] I have updated the CHANGELOG with a summary of my changes
- [ ] My changes have tests around them

PR #1496, opened 2020-08-28
Update README.md regarding Styled Components supported version

hellojere:
I had several SSR issues when building the app with the latest `styled-components` 5.x.y installed. React Static documentation doesn't mention the version support, but looking at the Issues it's not compatible with the 5.x.y, so would be great to mention this on the documentation for people to avoid spending time on debugging. 

About the issue itself; for some reason, the CSS class names generated with `styled-components` 5.x.y installed clash with each other. This happens especially when extending components like so: 

```JS
const MyElement = styled.div`
  color: pink;
`;

const OtherElement = styled(MyElement)`
  background-color: yellow;
`;
```

### DIFF
--- a/packages/react-static-plugin-styled-components/README.md
+++ b/packages/react-static-plugin-styled-components/README.md
@@ -17,3 +17,6 @@ export default {
   plugins: ["react-static-plugin-styled-components"]
 };
 ```
+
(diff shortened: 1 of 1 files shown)


SleeplessByte:
See https://github.com/react-static/react-static/pull/1473

Issue #1497, opened 2020-09-02
[Bug] Does not work with preact on production


rosskhanas:
### Reporting a bug?

![image](https://user-images.githubusercontent.com/10609041/91981135-0ae23a00-ed31-11ea-97d5-e72b62e3d676.png)

After investigation, I found that https://github.com/react-static/react-static/blob/87ea29379f4a8a73612db198c0bc5c021ba60366/packages/react-static/src/bootstrapApp.js#L19 line crashes.

### Environment

Run and copy the result of:

```
System:
    OS: macOS 10.15.5
    CPU: (4) x64 Intel(R) Core(TM) i7-5557U CPU @ 3.10GHz
    Memory: 2.27 GB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 13.12.0 - /usr/local/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.14.6 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Browsers:
    Chrome: 85.0.4183.83
    Safari: 13.1.1
  npmPackages:
    react-static: ^7.4.2 => 7.4.2
```

### Steps to Reproduce the problem

1.  Add preact aliases
2.  `yarn react-static build`
3. `yarn serve ./dist -p 3000`
4. Open `http://localhost:3000/` and the browser console.

### Expected Behavior

No JS crash

SleeplessByte:
Hi there. On `master` there is a [specific change](https://github.com/react-static/react-static/blob/f93fa5861c362316ea49f643cf1c95da7c9f1da8/CHANGELOG.md) for `preact`, referring to https://github.com/react-static/react-static/pull/1486. I think @tbgse has gotten it to work.

tbgse:
Are you using the `react-static-plugin-preact` and have followed the instructions in the [readme](https://github.com/react-static/react-static/tree/master/packages/react-static-plugin-preact)?

rosskhanas:
@tbgse with `react-static-plugin-preact` build completely fails, I guess it is because it uses preact 8 and it does not support hooks. That is why I tried to add preact 10 manually + applied changes manually to the `node.api.js` + removed `AppContainer` usage.

<img width="722" alt="image" src="https://user-images.githubusercontent.com/10609041/92008589-08dba380-ed50-11ea-8a05-a2a3c937d57d.png">

tbgse:
[...]

PR #1500, opened 2020-09-04
Move out Suspense to a separate file

rosskhanas:
Moved out Suspense to a separate file as it was mentioned in https://github.com/react-static/react-static/issues/1497#issuecomment-686178587.

## Description

I don't mind improve this in the future, but with this change, it is possible to `import { Suspense } from "react-static"` and use it instead of React Suspense while using `preact`.

## Changes/Tasks

The idea is to be able to avoid stubbing `React.default.Suspense` while using `react-static` with `preact`. Even though it is not fully done, thanks to this PR it would be possible to `import { Suspense } from "react-static"` and remove setting `React.default.Suspense` with something like `patch-package`(at least as a temp solution).

## Motivation and Context

Preact 10 support in `react-static`.

## Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 - Update preact to latest version and removed outdated compat dependency ([#1486](https://github.com/react-static/react-static/pull/1486))
 - Add prefresh to replace react hot loader when using preact plugin ([#1486](https://github.com/react-static/react-static/pull/1486))
+- Add an instruction how to make it work with preact 10 ([#1497](https://github.com/react-static/react-static/pull/1500))
 
 
 ### Bugfix
--- a/packages/react-static/src/Suspense.js
+++ b/packages/react-static/src/Suspense.js
@@ -1,0 +1,13 @@
+import * as React from 'react'
+
+function Suspense({ key, children, ...rest }) {
+  return typeof document !== 'undefined' ? (
+    <React.Suspense key={key} {...rest}>
+      {children}
+    </React.Suspense>
+  ) : (
+    <React.Fragment key={key}>{children}</React.Fragment>
+  )
+}
+
+export default Suspense;
--- a/packages/react-static/src/bootstrapApp.js
+++ b/packages/react-static/src/bootstrapApp.js
@@ -4,6 +4,9 @@ import { staticInfoContext } from './browser/hooks/useStaticInfo'
 import Suspense from './Suspense'
 
 // Override the suspense module to be our own
+// This is expected to break when using preact
+// In order to make it work with preact 10, use `patch-package` to remove those 2 lines
+// Reference: https://github.com/react-static/react-static/pull/1500
 React.Suspense = Suspense
 React.default.Suspense = Suspense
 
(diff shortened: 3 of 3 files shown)


SleeplessByte:
This is good for `preact`, but it's a breaking change.

Potentially we can still fix the `React.default.Suspense` in case it can be edited? So basically it will not change anything for `react`, and work with `preact`, and then it's not a breaking change!

rosskhanas:
@SleeplessByte sorry, missed your comment. Why is it breaking change? Just moved a single component out to a separate file...

SleeplessByte:
Before, a non preact user could import Suspense from react. Now they need to import it from react-static. If they don't, their SSR (build step) will fail. That makes it a breaking change.

rosskhanas:
@SleeplessByte in this PR there is no suspense stub removal, so users can still import from React.

SleeplessByte:
No, the fix to turn for Suspense during SSR has been removed for the regular import.

rosskhanas:
But it is still there, no?

image

SleeplessByte:
Doesn't show up like that in my diff 🤷 .

I guess that's okay then, but doesn't...

https://github.com/react-static/react-static/pull/1500/files#diff-f97a38894298c1e954f22b950cd27c6fL18-L19

...still break with `preact`?

rosskhanas:
@SleeplessByte it breaks, but the idea is to use something like `patch-package` as a hack to remove Suspense stubs in case `preact` is used.

```
React.Suspense = Suspense
React.default.Suspense = Suspense
```

I understand it would be great to get rid of those 2 lines on preact somehow nicer, but I could not find a solution so there are no breaking changes for react 😢 

If you feel like it does not make sense to merge this because it does not fully solve the problem, I don't mind closing the PR :)

SleeplessByte:
No, I think this is a good _first step_. Just trying to understand what is done and what still needs to be done. Thank you for this!

Alright, then I have 2 more requests:

1. Add a comment above these lines referring back to this PR (url) with something like "This is expected to break when using preact". We can then create a `patch-package` diff and post it here, and that way, people have instructions how to work around it.
2. Add an entry to CHANGELOG.md

I'll then be happy to merge 💯

rosskhanas:
I made another commit. Here is a diff I generated:

```
--- bootstrapApp.js	2020-09-19 12:10:56.000000000 +0300
+++ bootstrapApp.preact.js	2020-09-19 12:14:23.000000000 +0300
@@ -1,14 +1,6 @@
 /* eslint-disable import/no-dynamic-require */
 import * as React from 'react'
 import { staticInfoContext } from './browser/hooks/useStaticInfo'
-import Suspense from './Suspense'
-
-// Override the suspense module to be our own
-// This is expected to break when using preact
-// In order to make it work with preact 10, use `patch-package` to remove those 2 lines
-// Reference: https://github.com/react-static/react-static/pull/1500
-React.Suspense = Suspense
-React.default.Suspense = Suspense
 
 const App = require(`${process.env.REACT_STATIC_ENTRY_PATH}`).default
 

```

@SleeplessByte let me know if I need to make any improvements 🙂

SleeplessByte:
Thank you for your contribution @rosskhanas. It was [just released](https://github.com/react-static/react-static/releases/tag/v7.5.0).

Issue #1504, opened 2020-09-10
[Bug] Misconfigured HMR option for extract-css-chunks-webpack-plugin


IanChen83:
### Reporting a bug?

In recent versions of `extract-css-chunks-webpack-plugin` (i.e. [v4.7.4](https://github.com/faceyspacey/extract-css-chunks-webpack-plugin/blob/v4.7.4/src/loader.js#L216)), [it uses `options.hmr` to configure hot-reload](https://github.com/faceyspacey/extract-css-chunks-webpack-plugin/blob/v4.7.4/src/loader-options.json#L18-L20). However, in https://github.com/react-static/react-static/blob/f93fa5861c362316ea49f643cf1c95da7c9f1da8/packages/react-static/src/static/webpack/rules/cssLoader.js#L46, it uses `options.hot`, which makes style reloading fail to work when used with Linaria. After using `node.api.js` to update the options, hot-reload started to work.

```
export default (/* pluginOptions */) => ({
  webpack: (config, { defaultLoaders }) => {
    defaultLoaders.cssLoader.loader[0].options.hmr = defaultLoaders.cssLoader.loader[0].options.hot
    return config
  },
})
```

### Environment

```
  System:
    OS: macOS 10.15.6
    CPU: (8) x64 Intel(R) Core(TM) i7-8559U CPU @ 2.70GHz
    Memory: 538.83 MB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 14.8.0 - /usr/local/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.14.7 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Browsers:
    Chrome: 85.0.4183.102
    Firefox: 80.0.1
    Safari: 13.1.2
  npmPackages:
    react: ^16.13.1 => 16.13.1
    react-dom: ^16.13.1 => 16.13.1
    react-hot-loader: ^4.12.21 => 4.12.21
    react-static: ^7.4.2 => 7.4.2
    react-static-plugin-sitemap: ^7.4.2 => 7.4.2
```

Also,
```
npm ls extract-css-chunks-webpack-plugin linaria
app@ /project-path
├── linaria@1.3.3
└─┬ react-static@7.4.2
  └── extract-css-chunks-webpack-plugin@4.7.5
```

### Steps to Reproduce the problem

[...]

PR #1505, opened 2020-09-10
Update extract-css-chunks-webpack-plugin options for hot-reload

IanChen83:
## Description

This PR adds `hmr` when configuring `extract-css-chunks-webpack-plugin` for hot-reloading. See #1504 for details.

## Changes/Tasks

- [x] Changed code

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 - Fix basepath configuration for React Router ([#1520](https://github.com/react-static/react-static/pull/1520))
 - Fix mismatched chunk names between bundle and export ([#1518](https://github.com/react-static/react-static/pull/1518))
 - Bump `git-promise` to 1.0.0, fixing a security vulnerability ([#1522](https://github.com/react-static/react-static/pull/1522))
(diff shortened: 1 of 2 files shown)


SleeplessByte (commented):
Should this not remove `hot: true`? Since it now uses `hmr`? Or am I missing something.

Can you also add the CHANGELOG.md entry?

IanChen83:
> Should this not remove `hot: true`? Since it now uses `hmr`? Or am I missing something.

What I understand is that [at one time (v4.6.0)](https://github.com/faceyspacey/extract-css-chunks-webpack-plugin/blob/v4.6.0/src/loader.js) `hot: true` is an internal API of the plugin (which is why it is used at first place). Removing the old flag might break backward compatibility of old versions of the plugin. But if upgrading the dependency is what we wanna go with, then removing the field makes sense.

Note that the version specified in package.json is `^4.6.0`

SleeplessByte:
I'm fine with upgrading! Let's merge this first and then we'll see.

SleeplessByte (approved):

SleeplessByte:
Thank you for your contribution @IanChen83. It was [just released](https://github.com/react-static/react-static/releases/tag/v7.5.0).

Issue #1211, opened 2019-05-31
[FLUSH CHUNKS]: Unable to find src/pages/PageManager in Webpack chunks.


ghost:
Hey guys,

Lately I've been encountering a weird log during build :

```
[FLUSH CHUNKS]: Unable to find src/pages/PageManager in Webpack chunks.
```

It shows pretty much after every template exported :

```
[FLUSH CHUNKS]: Unable to find src/pages/PageManager in Webpack chunks. Please check usage of Babel plugin.
Time to render to string and extract { '$$typeof': Symbol(react.element),
  type: [Function: FinalComp],
  key: null,
  ref: null,
  props: {},
  _owner: null }
[FLUSH CHUNKS]: Unable to find src/pages/PageManager in Webpack chunks. Please check usage of Babel plugin.
[FLUSH CHUNKS]: Unable to find src/pages/PageManager in Webpack chunks. Please check usage of Babel plugin.
[FLUSH CHUNKS]: Unable to find src/pages/PageManager in Webpack chunks. Please check usage of Babel plugin.
Time to render to string and extract { '$$typeof': Symbol(react.element),
  type: [Function: FinalComp],
  key: null,
  ref: null,
  props: {},
  _owner: null }
[FLUSH CHUNKS]: Unable to find src/pages/PageManager in Webpack chunks. Please check usage of Babel plugin.
[FLUSH CHUNKS]: Unable to find src/pages/PageManager in Webpack chunks. Please check usage of Babel plugin.
Time to render to string and extract { '$$typeof': Symbol(react.element),
  type: [Function: FinalComp],
  key: null,
  ref: null,
  props: {},
  _owner: null }
[FLUSH CHUNKS]: Unable to find src/pages/PageManager in Webpack chunks. Please check usage of Babel plugin.
```

It doesn't seem to affect the build, and the site works fine (even without JS on so html is being rendered correctly).

Any idea on what's causing this ?

nddery:
Unsure how I missed this issue but I just posted in the Spectrum chat about what seem quite similar: https://spectrum.chat/react-static/general/flush-chunks-warnings-errors-when-building-outside-of-react-static-root~311c52c1-bd17-45a1-b4d1-376337072681

Do you also modify the `config.paths.dist` config ?

simonunlikely:
Hey,

I ran into your post I think, seems like the same problem but I did not modify the `config.paths.dist`.

[...]

PR #1518, opened 2020-10-11
Fix mismatched chunk names

chucklam:
# Mismatched chunk names between bundle and export

## Description
When the output directory (`config.paths.dist`) is changed to somewhere outside the working directory, there's a lot of warning messages during build.

To reproduce the problem:
```
react-static create --template basic
```

Run `yarn build` right away and it works fine.

Add to `static.config.js` a `paths.dist` property that points somewhere outside the working directory.
```
export default {
  paths: {
    dist: '/tmp/react-static'
  },
  getRoutes: ...,
  plugins: ...,
}
```

Run `yarn build` again and it gives a bunch of warnings.
```
Exporting HTML across 4 threads...
[FLUSH CHUNKS]: Unable to find __react_static_root__/src/pages/404 in Webpack chunks. Please check usage of Babel plugin.
[FLUSH CHUNKS]: Unable to find __react_static_root__/src/containers/Post in Webpack chunks. Please check usage of Babel plugin.
[FLUSH CHUNKS]: Unable to find __react_static_root__/src/pages/index in Webpack chunks. Please check usage of Babel plugin.
[FLUSH CHUNKS]: Unable to find __react_static_root__/src/containers/Post in Webpack chunks. Please check usage of Babel plugin.
[FLUSH CHUNKS]: Unable to find __react_static_root__/src/containers/Post in Webpack chunks. Please check usage of Babel plugin.
[FLUSH CHUNKS]: Unable to find __react_static_root__/src/containers/Post in Webpack chunks. Please check usage of Babel plugin.
[FLUSH CHUNKS]: Unable to find __react_static_root__/src/containers/Post in Webpack chunks. Please check usage of Babel plugin.
[FLUSH CHUNKS]: Unable to find __react_static_root__/src/containers/Post in Webpack chunks. Please check usage of Babel plugin.
...
```

## Motivation and Context

https://github.com/react-static/react-static/issues/1211 has an ongoing discussion on this.

In the bundling phase, templates are generated. And if `config.paths.dist` is different from "ROOT", the webpack chunk name for each template is [computed and assigned](https://github.com/react-static/react-static/blob/master/packages/react-static/src/static/generateTemplates.js#L36) with `chunkNameFromFile()`.

In the export phase, it's doing some pre-rendering magic with `react-universal-component` that I don't understand... But at one point it also references chunk names, and in the same case when `config.paths.dist` is different from "ROOT", it [calls](https://github.com/react-static/react-static/blob/master/packages/react-static/src/static/exportRoute.js#L150) `absoluteToRelativeChunkName()` to recompute the chunk name and truncate any prefix. There it looks asymmetric that `chunkNameFromFile()` is called when there's a prefix to truncate ([L36](https://github.com/react-static/react-static/blob/master/packages/react-static/src/utils/chunkBuilder.js#L36)) but not when there isn't ([L31](https://github.com/react-static/react-static/blob/master/packages/react-static/src/utils/chunkBuilder.js#L31)). I assume that's an unintentional omission as adding `chunkNameFromFile()` back in also fixes the problem above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - Fix eslintignore not ignorning package files ([#1486](https://github.com/react-static/react-static/pull/1486))
 - Remove `@types/react-hot-loader` from TypeScript template ([#1485](https://github.com/react-static/react-static/pull/1485))
 - Expand `styled-components` peer dependency version range in `react-static-plugin-styled-components` to allow newer versions of styled-components to be used ([#1473](https://github.com/react-static/react-static/pull/1473))
+- Fix mismatched chunk names between bundle and export ([#1518](https://github.com/react-static/react-static/pull/1518))
 
 
 ## 7.4.1
--- a/packages/react-static/src/utils/__tests__/chunkBuilder.test.js
+++ b/packages/react-static/src/utils/__tests__/chunkBuilder.test.js
@@ -53,5 +53,34 @@ describe('utils/chunkBuilder', () => {
         absoluteToRelativeChunkName('C:\\foo\\bar\\bazz\\', 'src-bar')
       ).toBe('src-bar')
     })
+    it('generates relative chunk names on absolute routes', () => {
+      expect(
+        absoluteToRelativeChunkName(
+          '/foo/bar/bazz/',
+          '/foo/bar/bazz/src/component.jsx'
+        )
+      ).toBe('src-component')
+    })
+
+    it('generates relative chunk names on absolute routes on windows', () => {
+      expect(
+        absoluteToRelativeChunkName(
+          'C:\\foo\\bar\\bazz\\',
+          'C:\\foo\\bar\\bazz\\src\\component.jsx'
+        )
+      ).toBe('src-component')
+    })
+
+    it('generates relative chunk names on relative routes', () => {
+      expect(absoluteToRelativeChunkName('/foo/bar/bazz/', 'src/bar')).toBe(
+        'src-bar'
+      )
+    })
+
+    it('generates relative chunk names on relative routes on windows', () => {
+      expect(
+        absoluteToRelativeChunkName('C:\\foo\\bar\\bazz\\', 'src\\bar')
+      ).toBe('src-bar')
+    })
   })
 })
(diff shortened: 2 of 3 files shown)


SleeplessByte:
I think you're right.

@nddery if you have time, can you confirm?

nddery:
@SleeplessByte that all make sense to me.

@chucklam thanks for looking into this! Might be worth adding a test for this ?

chucklam:
Tests added.

After looking at the existing tests, I also decided to call `chunkNameFromFile()` earlier in the code. It preserves the existing behaviors and makes `chunkName` and `pathPrefix` look more comparable.

SleeplessByte (approved):
Great. 

Can you add the entry to CHANGELOG.md?

SleeplessByte:
Looking great!

alqamabinsadiq:
@chucklam  I added your suggested fix in my react-static version, and it looks like it is not working. I still see the [FLUSH CHUNKS] error.

chucklam:
@alqamabinsadiq Actually I did it your hacky way in the beginning too :rofl: I went with the "proper" way only after I isolated the problem.

Based on your description, I suspect you didn't run `yarn build` in the `node_modules/react-static/` directory after making the changes. The actual code is being picked up at `node_modules/react-static/lib/` not `node_modules/react-static/src/`.

If that's not the problem, can you log `ROOT`, `pathPrefix`, and `chunkName` (both before and after `chunkNameFromFile`)? There may be cases I haven't tested.

alqamabinsadiq:
@chucklam Strange thing happened to me. I ran `npm install` again after deleting the node_modules, and it worked for me without even adding the above fix you mentioned. I can no longer see [FLUSH CHUNKS] errors. The main bug which I am facing is the reloading of content. After refreshing the page, my page is being rerendered. 

Check this out: https://experience-1605206018346.elphie.io/alqama/

Check, I think something is wrong with the routes. Just so you know the top navigation bar is top of routes that's why it is not re-rendering it. I am also prefetching the other routes info as you can see in the network tab.

SleeplessByte:
This was also included in 7.5.0.

PR #1551, opened 2020-12-26
[docs] Update Material UI guide for MUI v4

### LINKED ISSUE #1179: [Docs] Update Material UI Guide for v4
### Reporting a bug?

The guide in the repository doesn't work for the latest Material UI. API for SSR has changed. MUI v4 is still in Beta, but an update needs to be made. Any suggestions of how it can be patched temporarily before the official release is welcome too @tannerlinsley @sw-yx 

Latest API: https://next.material-ui.com/guides/server-rendering/
Oudated Guide: https://github.com/nozzle/react-static/blob/master/docs/guides/material-ui.md

cooperka:
Resolves #1179 as mentioned [here](https://github.com/react-static/react-static/issues/1179#issuecomment-751135591). The plugin was out of date and no longer worked on the latest version of MUI.

Note: This does not update the generic JSS plugin (released as a separate node module) which may or may not still work (I haven't tried it with anything other than MUI). Rather than doing the work of maintaining a separate plugin for MUI, I think the community would be better served keeping this simpler guide updated and lettings folks customize according to need.

Thanks to @Herohtar and @6pm earlier for their work debugging this.

### DIFF
--- a/docs/guides/material-ui.md
+++ b/docs/guides/material-ui.md
@@ -9,11 +9,10 @@ To use Material UI's server-side rendering (SSR) with React Static:
 ```js
 // your_app/plugins/jss-provider/node.api.js
 
-import { ServerStyleSheets } from '@material-ui/core';
+import { ServerStyleSheets } from '@material-ui/core/styles';
 
 export default () => ({
   beforeRenderToHtml: (App, { meta }) => {
-    // eslint-disable-next-line no-param-reassign
     meta.muiSheets = new ServerStyleSheets();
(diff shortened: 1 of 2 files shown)


SleeplessByte (approved):
Cool!

Have you also tried if these changes make your build work and make the plugin _actually_ work with Material UI? I think yes from your PR message, but I want to be certain before I merge.

cooperka:
Yes indeed, this works great for me now (the issues I mentioned turned out to be unrelated).

PR #1557, opened 2021-01-12
Make getComponentForPath properly return 404 page when route is invalid and 404 page exists

jkatofsky:
I was following the [animated routes guide](https://github.com/react-static/react-static/blob/master/docs/guides/animated-routes.md), and everything seemed to be working well. Then, I accidentally navigated to a non-existent route and realized that instead of my 404 page rendering automatically as it does when I don't use the `render` prop of `Routes`, nothing at all was rendered.

I dug into the implementation of `getComponentForPath` and realized that it always returns `false` if `routePath` does not match anything, even if a 404 template exists. This is due to [this line of code](https://github.com/react-static/react-static/blob/3793552f6e9ff921cb8480768a7438e386f6194b/packages/react-static/src/browser/components/Routes.js#L43) which should contain a `&&` instead of a `||` - and thus only return false when `routePath` 404s _and_ there is no `Comp` to render. This PR makes that change.

To fix things in the meantime, you can add this line of code to your custom `render` prop of `Routes`:

```JSX
//extra import to make it work
import { templatesByPath } from 'react-static';

<Routes
    render={({ routePath, getComponentForPath }) => {
    let element = getComponentForPath(routePath);

    //explicitly get the 404 page
    if (!element) element = React.createElement(templatesByPath['404']);

    //do the custom render...
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 - Fix mismatched chunk names between bundle and export ([#1518](https://github.com/react-static/react-static/pull/1518))
 - Bump `git-promise` to 1.0.0, fixing a security vulnerability ([#1522](https://github.com/react-static/react-static/pull/1522))
 - Fix misconfigured HMR option for extract-css-chunks-webpack-plugin ([#1505](https://github.com/react-static/react-static/pull/1505))
+- Make `getComponentForPath` properly return 404 page when route is invalid and 404 page exists ([#1557](https://github.com/react-static/react-static/pull/1557))
 
 ## 7.4.1
 
--- a/packages/react-static/src/browser/components/Routes.js
+++ b/packages/react-static/src/browser/components/Routes.js
@@ -40,7 +40,7 @@ function getTemplateForPath(path) {
  */
 function getComponentForPath(path) {
   const { Comp, is404 } = getTemplateForPath(path)
-  if (is404 || !Comp) {
+  if (is404 && !Comp) {
     return false
   }
 


SleeplessByte:
Nice. I think this is right.

- Can you add this to CHANGELOG.md
- Can you add a test for it failing without a 404 registered (should render default)?
- Can you add a test for it failing with a 404 registered (should render the 404 component with is404)?

jkatofsky:
@SleeplessByte Whoops - sorry for the trigger-happy PR. I added this to the changelog.

When it fails with no 404, the behaviour intended for `getComponentForPath` seems to be simply returning false. Otherwise, the 404 page should be returned with `is404`, as you said.

As for the tests, I don't see a test file in the `browser/components` directory. Should I start one?

SleeplessByte:
> When it fails with no 404, the behaviour intended for getComponentForPath seems to be simply returning false. Otherwise, the 404 page should be returned with is404, as you said.

Yeah. That makes sense to me. Does it make sense to you as a _user/developer_ using `react-static` ?

> As for the tests, I don't see a test file in the browser/components directory. Should I start one?

Right. Uhm. Yes, but if it's _very_ hard, you may do a manual test. This PR will serve as documentation that it was manually tested and works 🥇

jkatofsky:
> Does it make sense to you as a user/developer using react-static?

Yes, it makes sense to me. After this PR, I think `getComponentForPath` will behave exactly as `Routes` would without the custom rendering, which makes the most sense to me. Then, one can just focus on the actual rendering logic.

> Yes, but if it's very hard, you may do a manual test.

I've manually tested this on my project and it works (behaviour is as expected with or without a 404 template in the repository). I feel good about it, especially given the small scope of the change.

What I'm really trying to say is that I have little experience writing tests in JS and don't know how I'd approach this 😅. That being said, I'd be happy to do the testing (it's a good excuse to learn Jest) but I won't have the time to do so until later this week.

SleeplessByte:
> What I'm really trying to say is that I don't have a lot of experience writing tests in JS and at first glace I'm unsure how I'd approach this 😅. That being said, I'd be happy to buckle down and figure it out but I wouldn't have the time to do so until later this week.

All good. If you still want to do that, new PR is best! Manual testing is fine for this.

SleeplessByte:
Thank you for your contribution @jkatofsky . It was [just released](https://github.com/react-static/react-static/releases/tag/v7.5.0).

PR #1563, opened 2021-01-30
Upgrade postcss-loader to 4.2.0

jeremy-hanna:
## Description

Hi there! :wave: 

PostCSS 8.0 has been released for awhile and with it came some [breaking changes for plugin APIs](https://evilmartians.com/chronicles/postcss-8-plugin-migration) - it maintains backwards compatibility with older API as far as I can tell.

However the new plugin API wasn't supported in [postcss-loader until 4.0+](https://github.com/webpack-contrib/postcss-loader/blob/master/CHANGELOG.md#403-2020-10-02) but we're locked at 3.0, effectively limiting use of plugins that require the PostCSS 8.0 API

This PR upgrades the postcss-loader, fixes the breaking loader configuration, and adds PostCSS as a peerDependency.

## Changes/Tasks

- [x] Upgrade the postcss-loader and core rule
- [x] Update plugins that used the loader
- [x] `yarn build` core + plugins successfully
- [x] Test local production build with `yarn link`ed changes

## Motivation and Context

My project uses TailwindCSS which ([as of 2.0](https://tailwindcss.com/docs/upgrading-to-v2#install-tailwind-css-v2-0-and-post-css-8)) utilizes the new PostCSS 8 plugin API so I need to upgrade this dependency upstream but we're locked at 3.0 hence the PR.

## Types of changes

## Checklist:

### DIFF
--- a/packages/react-static/package.json
+++ b/packages/react-static/package.json
@@ -78,7 +78,7 @@
     "mutation-observer": "^1.0.3",
     "optimize-css-assets-webpack-plugin": "^5.0.3",
     "portfinder": "^1.0.21",
-    "postcss-flexbugs-fixes": "^5.1.0",
+    "postcss-flexbugs-fixes": "^4.1.0",
     "postcss-loader": "^4.2.0",
     "pretty-error": "^2.1.1",
     "progress": "^2.0.3",
--- a/packages/react-static/package.json
+++ b/packages/react-static/package.json
@@ -27,6 +27,7 @@
     "publishLink": "echo '{{event}} to {{changed}}' && yalc publish"
   },
   "peerDependencies": {
+    "postcss": "^8.2.4",
     "react-hot-loader": "^4.12.11"
   },
   "dependencies": {
@@ -75,8 +76,8 @@
     "mutation-observer": "^1.0.3",
     "optimize-css-assets-webpack-plugin": "^5.0.3",
     "portfinder": "^1.0.21",
-    "postcss-flexbugs-fixes": "^4.1.0",
-    "postcss-loader": "^3.0.0",
+    "postcss-flexbugs-fixes": "^5.1.0",
+    "postcss-loader": "^4.2.0",
     "pretty-error": "^2.1.1",
     "progress": "^2.0.3",
     "prop-types": "^15.7.2",
--- a/packages/react-static/package.json
+++ b/packages/react-static/package.json
@@ -27,7 +27,7 @@
     "publishLink": "echo '{{event}} to {{changed}}' && yalc publish"
   },
   "peerDependencies": {
-    "postcss": "^8.2.4",
+    "postcss": ">= 7.0.0, < 9.0.0",
     "react": "^16.9.0",
     "react-dom": "^16.9.0",
     "react-hot-loader": "^4.12.11"
--- a/packages/react-static/src/static/webpack/rules/cssLoader.js
+++ b/packages/react-static/src/static/webpack/rules/cssLoader.js
@@ -17,13 +17,14 @@ function initCSSLoader() {
         // Necessary for external CSS imports to work
         // https://github.com/facebookincubator/create-react-app/issues/2677
         sourceMap: true,
-        ident: 'postcss',
-        plugins: () => [
-          postcssFlexbugsFixes,
-          autoprefixer({
-            flexbox: 'no-2009', // I'd opt in for this - safari 9 & IE 10.
-          }),
-        ],
+        postcssOptions: {
+          plugins: [
+            postcssFlexbugsFixes,
+            autoprefixer({
+              flexbox: 'no-2009', // I'd opt in for this - safari 9 & IE 10.
+            }),
+          ],
+        },
       },
     },
   ]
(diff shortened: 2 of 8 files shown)


packages/react-static/src/static/webpack/rules/cssLoader.js
```diff
@@ -17,13 +17,14 @@ function initCSSLoader() {
         // Necessary for external CSS imports to work
         // https://github.com/facebookincubator/create-react-app/issues/2677
         sourceMap: true,
-        ident: 'postcss',
-        plugins: () => [
```
jeremy-hanna, line 21: This is a change with the postcss-loader 4.0 configuration API - this is able to be an array and the key is part of `postcssOptions`

https://github.com/webpack-contrib/postcss-loader/blob/435224d134f6193d42859988e047b6dbe8dcdbce/CHANGELOG.md#-breaking-changes

>>> PostCSS (plugins/syntax/parser/stringifier) options was moved to the postcssOptions option, please look at [docs](https://github.com/webpack-contrib/postcss-loader#postcssoptions)

packages/react-static/src/static/webpack/rules/cssLoader.js
```diff
@@ -17,13 +17,14 @@ function initCSSLoader() {
         // Necessary for external CSS imports to work
         // https://github.com/facebookincubator/create-react-app/issues/2677
         sourceMap: true,
-        ident: 'postcss',
```
jeremy-hanna, line 20: This is no longer needed

https://github.com/webpack-contrib/postcss-loader/blob/master/CHANGELOG.md#notes

> - you don't need ident option for loader

packages/react-static/src/utils/binHelper.js
```diff
@@ -116,15 +116,15 @@ ignoredExtensions.forEach(ext => {
 
 const originalConsoleError = console.error
 console.error = (...args) => {
-  if(args.length === 0){
+  if (args.length === 0) {
```
jeremy-hanna, line 119: I think this is just my prettier config - I can back this change out if you want

jeremy-hanna:
Of note, postcss-loader 4.0 dropped support for Webpack <4 (but we're locked above that) and node <10 - I think these are fine but wasn't sure what the support requirements were here

https://github.com/webpack-contrib/postcss-loader/blob/master/CHANGELOG.md#-breaking-changes

> warning BREAKING CHANGES
>
>    minimum supported Node.js version is 10.13
>    minimum supported webpack version is 4

packages/react-static/src/static/webpack/rules/cssLoader.js
jeremy-hanna: This also technically breaks updates to the `node.api.js` if the user attempts to find and append to the plugins as there is another object to dig through

eg.
```
  const plugins = postcssLoader.options.plugins;

  Object.assign(postcssLoader.options, { 
    plugins: [...new_plugins, plugins]
  }
```

must be changed to
```
  const plugins = postcssLoader.options.postcssOptions.plugins;
  
  Object.assign(postcssLoader.options.postcssOptions, {
    plugins: [...new_plugins, plugins]
  } 
```

I'm not entirely sure what the community conventions around "breaking changes" referring to webpack config is - but that might push this to a minor version upgrade instead of a patch :shrug:

packages/react-static/package.json
```diff
@@ -27,6 +27,7 @@
     "publishLink": "echo '{{event}} to {{changed}}' && yalc publish"
   },
   "peerDependencies": {
+    "postcss": "^8.2.4",
```
jeremy-hanna, line 30: this may be more involved - looks like several other deps depend on ^7.0.0 such as the css-loader :disappointed:

packages/react-static/src/static/webpack/rules/cssLoader.js
SleeplessByte: If it's a breaking change, it _should_ be a major bump but 🤷🏽‍♂️

packages/react-static/package.json
SleeplessByte: There "just" peer-dependencies tho. It's okay to upload it accross the board, or do it incrementally. You can change it to `>= 7.0.0, < 9.0.0` for example.

packages/react-static/src/static/webpack/rules/cssLoader.js
jeremy-hanna: yeah, :point_up: is the equivalent of monkey patching a runtime method in Ruby for webpack plugins imo but I'm not sure if anyone does that

packages/react-static/package.json
jeremy-hanna: that is a good point, with the upgrade of postcss-loader incrementally getting the other deps are easier and this is only upgrading the loader :slightly_smiling_face:

packages/react-static/src/static/webpack/rules/cssLoader.js
SleeplessByte: Eh. This PR is linked in the CHANGELOG, including this conversation 😋

SleeplessByte:
Can you rebase please? I think this is ready to go :)

ivanoats:
PostCSS loader is now at 6.2.1. How should we handle this PR now - update it? make a new one? Happy to help.

ivanoats:
@jeremy-hanna @SleeplessByte Hi folks - I'm happy to help on this one. Just let me know which direction to go - rebase or new PR with updated version?

Issue #1568, opened 2021-02-11
[Bug] assetsPath and path.assets break tempate loading


codejunkienick:
### Reporting a bug?

When adding `assetsPath: 'assets'` and `paths: { assets: 'assets' }` to basic template it breaks page with message
`An error occurred loading this page's template. More information is available in the console.`

console.logs contain this error

```
main.7d6f7928.js:1 ChunkLoadError: Loading chunk 4 failed.
(error: http://localhost:3001/templates/__react_static_root__/src/pages/index.62bb425b.js)
    at Function.c.e (http://localhost:3001/assets/main.7d6f7928.js:1:1197)
    at load (http://localhost:3001/assets/main.7d6f7928.js:1:4402)
    at http://localhost:3001/assets/templates/vendors~main.8117e77c.js:1:121425
    at new Promise (<anonymous>)
    at requireAsync (http://localhost:3001/assets/templates/vendors~main.8117e77c.js:1:121123)
    at n.value (http://localhost:3001/assets/templates/vendors~main.8117e77c.js:1:7511)
    at n.value (http://localhost:3001/assets/templates/vendors~main.8117e77c.js:1:8631)
    at new n (http://localhost:3001/assets/templates/vendors~main.8117e77c.js:1:7339)
    at wi (http://localhost:3001/assets/templates/vendors~main.8117e77c.js:1:177755)```

Actual path to template is `http://localhost:3001/assets/templates/__react_static_root__/src/pages/index.62bb425b.js`

### Environment

```
  System:
    OS: Linux 5.10 Manjaro Linux
    CPU: (16) x64 AMD Ryzen 7 3800X 8-Core Processor
    Memory: 9.98 GB / 31.33 GB
    Container: Yes
    Shell: 3.1.2 - /usr/bin/fish
  Binaries:
    Node: 15.8.0 - /usr/bin/node
    Yarn: 1.22.10 - /usr/bin/yarn
    npm: 6.14.11 - /usr/bin/npm
    Watchman: 4.9.0 - /usr/bin/watchman
  Browsers:
    Firefox: 85.0.1
  npmPackages:
    react: ^16.9.0 => 16.14.0 
    react-dom: ^16.9.0 => 16.14.0 
    react-hot-loader: ^4.13.0 => 4.13.0 
    react-static: 7.4.2 => 7.4.2 
    react-static-plugin-reach-router: 7.4.2 => 7.4.2 
    react-static-plugin-sitemap: 7.4.2 => 7.4.2 
[...]

PR #1569, opened 2021-02-12
REACT_STATIC_ASSETS_PATH as webpack.output.publicPath

codejunkienick:
## Description

webpack.output.publicPath for production deployment should use assetsPath value

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,6 @@
 ### Improved
 
 ### Bugfix
-- Make sure assetsPath is used for webpack build and template generation
 
 ## 7.5.0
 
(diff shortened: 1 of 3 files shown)


SleeplessByte:
_Why_ should it use asset path? We need to explain this to others.

codejunkienick:
> _Why_ should it use asset path? We need to explain this to others.

It already uses assetsPath param in dev configuration. 

Basically if we do not set webpack publicPath to assetsPath parameter, then all links to the templates/{template_name}.js will be linked to the basePath or root, resulting in page crash.

Here is linked bug
https://github.com/react-static/react-static/issues/1568

SleeplessByte:
Great. That makes sense. Yes this could have been a typo.

Can you add the PR number and link to the CHANGELOG, like everything else? I'll merge it after.

SleeplessByte:
Thank you @codejunkienick :)

Issue #1566, opened 2021-02-10
routeInfo hook


andreme:
**Is your feature request related to a problem? Please describe.**
I'm trying to develop a solution to transform images (like https://github.com/react-static/react-static/issues/1130 ).
But I would like to make use of the existing multithreading and convert them in the worker process.
For that I need to write some data back into into the routeInfo, and as far as I can see it needs to be done before it is used anywhere else: https://github.com/react-static/react-static/blob/58cf1c6a3b2847944b588589b8abccc3a2f083c9/packages/react-static/src/static/exportRoute.js#L107

**Describe the solution you'd like**
If I code a PR for that, will it be accepted?

SleeplessByte:
It was merged!

Released as 7.5.3

PR #1572, opened 2021-02-19
Allow plugins to modify routeInfo

andreme:
## Description

Allow plugins to modify routeInfo to for example add or modify the data that is used on a page.

## Changes/Tasks

- Call new routeInfo hook in exportRoute
- Call new routeInfo hook in dev server for each request
- Extract valid hooks from object, to remove extra list

## Motivation and Context

#1566

## Types of changes

## Checklist:

### DIFF
--- a/packages/react-static/src/static/plugins.js
+++ b/packages/react-static/src/static/plugins.js
@@ -1,39 +1,6 @@
 import { getHooks, reduceHooks } from '../utils'
 
-const supportedHooks = [
-  'afterGetConfig',
-  'beforePrepareBrowserPlugins',
-  'afterPrepareBrowserPlugins',
-  'beforePrepareRoutes',
-  'getRoutes',
-  'normalizeRoute',
-  'afterPrepareRoutes',
-  'webpack',
-  'afterBundle',
-  'afterDevServerStart',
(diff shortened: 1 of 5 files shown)


packages/react-static/src/static/plugins.js
```diff
@@ -1,39 +1,6 @@
 import { getHooks, reduceHooks } from '../utils'
 
-const supportedHooks = [
```
SleeplessByte, line 3: What's the reason these ae removed? Even though they don't have a default implementation, are you certain they're not actually supported anymore?

andreme: All the entries in the array are also keys in the exported object. So I just removed the duplication. validatePlugin now takes the valid hook names from the exported object.

SleeplessByte: GitHub was collapsing those lines, so I didn't realise. This is good. Thanks :)

SleeplessByte (approved):

SleeplessByte:
(It's still a draft, but overall this is looking good to me)

andreme:
Thanks @SleeplessByte

Issue #1589, opened 2021-03-21
[Bug] TypeScript template does not support image import


ngocdaothanh:
I'd like to import an image, but there's error:

```
TS2307: Cannot find module './logo.png' or its corresponding type declarations.
     6 | import './app.css'
     7 |
  >  8 | import logo from './logo.png'
       |                  ^^^^^^^^^^^^
```

How to fix please?

ngocdaothanh:
I like the simplicity of react-static very much, but because of this issue, I have to switch to Gatsby (hopefully just temporarily).

ksekachev:
@ngocdaothanh hi, I think this issue is not related to React-static, please, try to follow Step 3 and Step 4 from [this guide](https://stackoverflow.com/questions/43638454/webpack-typescript-image-import?rq=1).

SleeplessByte:
@ksekachev is correct. This is unrelated.

ngocdaothanh:
It's not related to react-static core, but the TypeScript template is inside react-static repo, it should be improved to have the image import preconfigured, so that users won't have trouble using react-static?

https://github.com/react-static/react-static/tree/master/packages/react-static/templates/typescript

SleeplessByte:
Yes, we'll definitely take a PR 🥳 😄

ngocdaothanh:
👍 I've created this PR:
https://github.com/react-static/react-static/pull/1597

PR #1597, opened 2021-04-07
Add image import example for TypeScript template

ngocdaothanh:
See https://github.com/react-static/react-static/issues/1589


packages/react-static/templates/typescript/index.d.ts
```diff
@@ -0,0 +1,4 @@
+// Allow importing images,
+// you can add other formats supported by file-loader
+declare module '*.jpg'
```
SleeplessByte, line 3: Perhaps we want to copy the type definitions from create-react-app?

ngocdaothanh: Would you please provide the link to that?

I've tried looking at its repo, but couldn't find anything interesting:
https://github.com/facebook/create-react-app

PR #1599, opened 2021-04-11
Show a warning in development mode, when a template file does not have a default export

andreme:
At the moment, there is no indication why the 404 page is displayed instead of the actual page


SleeplessByte (approved):
Sounds like a good addition. Thanks!

PR #1608, opened 2021-05-08
Pass full route and state to routeInfo hook

andreme:
Found some issues with the routeInfo hook I'd added:
- not all the route data was passed to it
- state was not passed through


andreme:
@SleeplessByte I just realised that the hook was storing all route properties.
That's fixed now, only the properties that were stored before plus the props the hook adds are stored.

SleeplessByte:
(re-opened to retrigger CI)

